PR #9, opened 2018-05-11
Add handling of readiness annotation

weimerb:


src/property/FaceliftProperty.h
```diff
@@ -77,6 +85,19 @@ class PropertyBase
         return m_ownerSignal;
     }
 
+    bool &readiness()
```
jacky309: I feel like isReady() would be a better name

src/property/FaceliftProperty.h
```diff
@@ -77,6 +85,19 @@ class PropertyBase
...
+    {
+        return m_readiness;
+    }
+
+    void setReadiness(bool ready)
```
jacky309: setReady ?

src/property/FaceliftProperty.h
jacky309: and it should be const

weimerb: That might be rather confusing, if in all other places (even as annotation) we use "readiness" and only here "isReady". I tend to leave it as it is.

A non-const reference is returned here, so the function cannot be made const.

PR #12, opened 2018-05-23
Use relative filepath in facelift_add_qml_plugin_qmldir

weimerb:
QtCreator does not recognize absolute file paths in qmldir,
after this commit intellisense works in QtCreator for qmlfiles


weimerb:
Jacques, can you verify this on your Mac, as well?

PR #18, opened 2018-06-19
Modelipc4

weimerb:


codegen/facelift/templates/ServiceIPC.template.h
```diff
@@ -226,18 +259,31 @@ class {{interface}}IPCProxy : public facelift::IPCProxy<{{interface}}PropertyAda
...
-        {% if property.type.is_model %}
-        qFatal("Model not supported");
-        {% elif property.type.is_interface %}
+        {% if property.type.is_interface %}
         // qFatal("Property of interface type not supported");
         qWarning() << "TODO: handle interface properties";
+        {% elif property.type.is_model %}
+        int {{property.name}}Size;
+        msg >> {{property.name}}Size;
+        if ({{property.name}}Size && !m_{{property.name}}.size()) {
```
jacky309: As soon as the connection to the server is lost, we need to clear the local model, since we are not going to be able to fetch any new item from the server. Therefore, this test does not make sense since "m_{{property.name}}.size()" is 0.

PR #21, opened 2018-06-25
Use prefetching for models over IPC

weimerb:
Fetch several elements from server in one IPC message instead of only
the one requested.


jacky309:
Approved

PR #22, opened 2018-06-25
Document how to implement models

weimerb:


doc/page-implementation.h
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+private:
+    QStrigList m_items;
+};
+\endcode
+
+The property adapter will include a public member variable that can be used to provide the model
+data. In our case IModelPropertyAdapter includes \c m_theModel. Two methods of this member variable
+should be called (usually in the constructor): \c setSize and \c setGetter. \c setSize needs to be
```
jacky309: it sounds like setSize would have to be called somewhere by the user code, which is not the case since it is called only by Facelift itself

jacky309:
approved

PR #28, opened 2018-07-11
Clean up code

weimerb:
* Removed LGPL license file
* Added explicit 'emit' to signals
* Added Q_INTERFACES macro
* Updated .gitignore file


.gitignore
```diff
@@ -1 +1,37 @@
...
+
+CMakeFiles
 bin
+imports
+facelift_generated
+
+examples/launch-*.sh
+CMakeCache.txt
+doc/doxygen.cfg
+test/test-property
```
jacky309: I do not like to push such a file in our git repo.
Please use shadow builds.

weimerb: Why should this be enforced? Having the possibility should not bother others. We don't enforce usage of QtCreator instead of Eclipse either ;-)

jacky309: It bothers people who do use shadow builds, since they are not able to ensure that their source folder is really clean without first deleting the ".gitignore" file.

weimerb: Those "people" can always do a "git clean -fxd" or additionally with "-n" to see what would be deleted. This is a rather dogmatic approach that is not enforced on any other project. If you have a look in qtbase for instance, you'll see more than 300 rules.

PR #29, opened 2018-07-11
Async method doc

jacky309:


doc/page-features.h
```diff
@@ -2,6 +2,52 @@
 
 \page Features
 
+
+\section async Asynchronous methods
+
+In addition to standard methods, which block the caller until the method execution is finished and provide an immediate output value, Facelift
```
weimerb: s/output/return? ... as used below, as well

doc/page-features.h
```diff
@@ -2,6 +2,52 @@
...
+supports asynchronous methods. Asynchronous methods have been introduced to support operations which potentially last a long time, or which are
+not able to produce an output value immediately.
```
weimerb: ... same here

doc/page-features.h
```diff
@@ -2,6 +2,52 @@
...
+interface MyInterface {
+
+    @async : true
+    int myAsyncMethod(int inputArgument);
+
+}
+\endcode
+
+\subsection async-sub2 Usage from QML UI code
+In order for the method's return value to be provided asynchronously, the method needs to be called with a javascript function object as argument, in addition to the
```
weimerb: Could you limit the line lengths a bit (I even used only 100 chars), so it can be read on laptop, as well?
s/javascript/JavaScript?

doc/page-features.h
```diff
@@ -2,6 +2,52 @@
...
+"normal" argument. The javascript function should accept an argument, unless the method is defined as returning "void".
```
weimerb: s/"normal" argument/input parameter?

doc/page-features.h
```diff
@@ -2,6 +2,52 @@
...
+\code
+    myInterface.myAsyncMethod(integerInputArgument, function(integerResult) {
+        print("myAsyncMethod finished execution with result " + integerResult);
+    });
+\endcode
+
+\subsection async-sub3 C++ method implementation
+In order for a C++ implementation to provide its output value asynchronously, an asynchronous method takes an additional argument. That argument defines
+a "operator(...)" method, which means it can be called as if it was a std::function. Although it can be called synchronously, that object would typically
```
weimerb: ellipses are confusing - it is operator()
s/defines a "operator(...)" method/is a functor (it defines operator()) ?
s/as if it was a std::function/like a function ?

PR #30, opened 2018-07-13
Immediately update properties over IPC

weimerb:
All properties of an interface will be synchronized after a method or
property setter call now.

### DIFF
--- a/src/ipc/dbus/ipc-dbus.h
+++ b/src/ipc/dbus/ipc-dbus.h
@@ -798,7 +798,7 @@ public:
     }
 
     template<typename PropertyType>
-    void sendSetterCall(const char *methodName, const PropertyType &value)
+    DBusIPCMessage sendSetterCall(const char *methodName, const PropertyType &value)
     {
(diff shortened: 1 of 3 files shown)


src/ipc/dbus/ipc-dbus.h
```diff
@@ -894,19 +895,35 @@ class DBusIPCProxy : public IPCProxyBase<AdapterType, IPCAdapterType>, protected
...
+        }
     }
 
     template<typename ReturnType, typename ... Args>
     void sendMethodCallWithReturn(const char *methodName, ReturnType &returnValue, const Args & ... args)
     {
-        auto msg = m_ipcBinder.sendMethodCall(methodName, args ...);
+        DBusIPCMessage msg = m_ipcBinder.sendMethodCall(methodName, args ...);
+        if (msg.isReplyMessage()) {
+            IPCTypeHandler<ReturnType>::read(msg, returnValue);
```
jacky309: msg >> returnValue ?

PR #34, opened 2018-07-20
Add auto tests

weimerb:
Inital auto tests based on QtTest have been added. The frontend (UI)
facing interface is tested. There are six test combinations: local,
via ipc and inprocess (server and client in one process), all three
with C++ and QML backend.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,8 @@ option(FACELIFT_ENABLE_CODEGEN "Enable code generator build and install. Should 
 option(FACELIFT_ENABLE_DBUS_IPC "Force DBus IPC support (no autodetection)" OFF)
 option(FACELIFT_DISABLE_DBUS_IPC "Force disable DBus IPC support (no autodetection)" OFF)
 option(FACELIFT_BUILD_EXAMPLES "Enable build of examples" ON)
-option(FACELIFT_BUILD_TEST "Enable build of tests" OFF)
+option(FACELIFT_BUILD_TESTS "Enable build of tests" OFF)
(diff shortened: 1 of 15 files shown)


cmake/faceliftMacros.cmake
```diff
@@ -607,6 +607,10 @@ function(facelift_add_qml_plugin PLUGIN_NAME)
         set(ARGUMENT_VERSION "1.0")
     endif()
 
+    if(NOT ARGUMENT_OUTPUT_BASE_DIRECTORY)
+        set(ARGUMENT_OUTPUT_BASE_DIRECTORY "import")
```
jacky309: imports ?

CMakeLists.txt
```diff
@@ -16,7 +16,8 @@ option(FACELIFT_ENABLE_CODEGEN "Enable code generator build and install. Should
 option(FACELIFT_ENABLE_DBUS_IPC "Force DBus IPC support (no autodetection)" OFF)
 option(FACELIFT_DISABLE_DBUS_IPC "Force disable DBus IPC support (no autodetection)" OFF)
 option(FACELIFT_BUILD_EXAMPLES "Enable build of examples" ON)
-option(FACELIFT_BUILD_TEST "Enable build of tests" OFF)
+option(FACELIFT_BUILD_TESTS "Enable build of tests" OFF)
+option(FACELIFT_BUILD_OBSOLETE_TEST "Enable build of obsolete tests" OFF)
```
jacky309: I would prefer to keep all tests under one option. We can remove any obsolete test later

PR #49, opened 2018-08-23
Windows fixes

jacky309:


codegen/facelift-codegen.py
```diff
@@ -225,6 +236,10 @@ def requiredInclude(self):
 
 setattr(qface.idl.domain.Module, 'namespaceCppOpen', property(namespaceCppOpen))
 setattr(qface.idl.domain.Module, 'namespaceCppClose', property(namespaceCppClose))
+#setattr(qface.idl.domain.Module, 'classExportDefines', classExportDefines)
+#setattr(qface.idl.domain.Module, 'classExport', classExport)
```
weimerb: If that is really not needed it should be removed.

jacky309: done

PR #51, opened 2018-08-28
QVariant support

JanKarasinski:


jacky309:
Supporting "var" type, and especially mapping it to QVariant, is going to cause a number of issues, since a QVariant object can store all sorts of data types which we are not able to make available over IPC. We might think about introducing some kind of variant type which can only store a limited number of simple data types, but we are not willing to introduce QVariant in any of Facelift's APIs.

jacky309:
Would such a variant type work for you ?

JanKarasinski:
Yes, it would. I wish i could create such variant, but i'm afraid i don't have enough time. Maybe i'll try something in my spare time

PR #52, opened 2018-08-30
Remove warnings

jacky309:


src/model/FaceliftModel.h
```diff
@@ -977,11 +977,15 @@ struct TypeHandler<Type *, typename std::enable_if<std::is_base_of<InterfaceBase
 
     static void write(BinarySeralizer &msg, const Type &param)
     {
+    	Q_UNUSED(msg);
```
weimerb: Indentation (below, as well)

PR #73, opened 2018-09-28
IPC: Use asynchronous object registry

jacky309:


src/ipc/dbus/ipc-dbus-object-registry.cpp
```diff
@@ -31,5 +31,96 @@
...
+    }
+}
+
+void DBusObjectRegistry::MasterImpl::init()
+{
+    m_objectRegistryAdapter.setService(this);
+    m_objectRegistryAdapter.init();
+}
+
+bool DBusObjectRegistry::MasterImpl::registerObject(QString objectPath, QString serviceName)
```
weimerb: any reason why parameters are not passed as const ref, as usual? Here and below (I know it is basically copy/paste).

jacky309: The reason is that this interface is generated from a QFace definition. All parameters are passed by value.

PR #95, opened 2018-10-31
add property binding backend for qml

rakeshsadhu:


tests/propertyBinding/impl/qml/PropertyBindingInterfaceQmlImplementation.qml
```diff
@@ -0,0 +1,68 @@
...
+
+    comboStr1 :cmb
+
+    ComboStructure {
+        id: cmb
+        iData: 100
+        sData: "binding"
+        bData : true
+    }
+
```
jacky309: function test () {
 cmb.iData = 1234;
TEST(comboStr1.iData == 1234);
}

rakeshsadhu: cmb.iData isn't accessible from testFunction , as its a private . So Instead of above code snippet I am changing it to 
as below in QMLBackend `Implementation


` 34 PropertyBindingInterfaceTestQMLImplementation {
 35     id: root
 36     Component.onCompleted: console.log("QML implementation is used.");
 37 
 38     comboStr2: comboStr1
 39 }
`

`  function test_intPropertyBinding() {
 39         api.comboStr1.iData = 99;
 40         compare(api.comboStr2.iData, 99);
 41     }
`

PR #104, opened 2018-11-13
add ci badges inside readme file

derived-coder:
1. This allows to see easily that this repo is using CI
2. Can easily jump to the CI servers via link
3. The status of these badges will change dynamically when the build breaks or runs through


derived-coder:
Done.

PR #112, opened 2018-11-22
Add tests for imported module

weimerb:


tests/combined/check_combined.js
```diff
@@ -53,8 +53,8 @@ function defaults() {
 
 function initialized() {
     api.initialize();
-
     tryVerify(function() { return api.boolProperty; });
```
jacky309: can we use the "ready" property instead of this "boolProperty" as a trigger ?

weimerb: No, because ready is already triggered in the defaults() function, where the default values are checked. We need this trigger to know when the initialization with different values is completed.

jacky309: Something more explicit than "boolProperty" would be better. It is a special boolean property

PR #128, opened 2018-12-21
Fixes

jacky309:


src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -0,0 +1,202 @@
+/**********************************************************************
+**
+** Copyright (C) 2018 Luxoft Sweden AB
```
weimerb: from now on 2019 in newly introduced files...

PR #130, opened 2019-01-03
Pass complex parameters in functions and signals as const ref

rakeshsadhu:


codegen/facelift/templates/IPCDBusProxy.template.cpp
```diff
@@ -126,7 +126,7 @@ void {{className}}::deserializeSignal(::facelift::dbus::DBusIPCMessage &msg)
 
 {% if operation.isAsync %}
 void {{className}}::{{operation.name}}(
-    {%- for parameter in operation.parameters -%}{{parameter.cppType}} {{parameter.name}}, {% endfor %}facelift::AsyncAnswer<{{operation.interfaceCppType}}> answer){% if operation.is_const %} const{% endif %} {
+    {%- for parameter in operation.parameters -%}{%if not parameter.type.is_primitive %}const {{parameter.cppType}} &{% else %}{{parameter.cppType}} {% endif %}{{parameter.name}}, {% endfor %}facelift::AsyncAnswer<{{operation.interfaceCppType}}> answer){% if operation.is_const %} const{% endif %} {
```
jacky309: We want to avoid code duplication.
I suggest you add a new property (alongside "cppType") called something like "cppMethodArgumentType" which returns either a const ref type or the actual type , depending on 
then you would replace this line by the following :
{%- for parameter in operation.parameters -%}{{parameter.cppMethodArgumentType}} {{parameter.name}}, {% endfor %}facelift::AsyncAnswer<{{operation.interfaceCppType}}> answer){% if operation.is_const %} const{% endif %} {

PR #136, opened 2019-01-23
Harmonize test and example classes names

rakeshsadhu:


examples/IPC/ui/AddressBookIPCServerLocalQMLModel.qml
```diff
@@ -57,7 +57,7 @@ Item {
         id: proxy
     }
 
-    AddressBookImplementationBase {
+    AddressBook {
```
weimerb: I don't think this is correct, because the implementation is provided here indeed and the implementation should be based on AddressBookImplementationBase.

examples/MyPackage/models/cpp/MyInterfaceCppImplementation.h
```diff
@@ -39,11 +39,11 @@ using namespace facelift::example::mypackage;
 /**
  * C++ Implementation of the MyInterface API
  */
-class MyInterfaceCppImplementation : public MyInterfaceImplementationBase
+class MyInterfaceImplementation : public MyInterfaceImplementationBase
```
weimerb: Please update the documentation accordingly (page-reference.h, etc.)

examples/IPC/ui/AddressBookIPCServerLocalQMLModel.qml
rakeshsadhu: yes you're right , its wrong  i mistyped it somehow .

rakeshsadhu:
@jacky309  kindly review and help me checking in .

examples/Tuner/models/qml/tuner/TunerViewModel.qml
```diff
@@ -37,7 +37,7 @@ import QtQuick.Controls 1.2
 /**
  * QML Implementation of the TunerViewModel API
  */
-TunerViewModelImplementationBase {
+TunerViewModel {
```
jacky309: no

rakeshsadhu: changed

PR #139, opened 2019-01-29
Make registration of uncreatable types explicit 

rakeshsadhu:


rakeshsadhu:
@weimerb  i have two pullrequests, kindly review https://github.com/Pelagicore/facelift/pull/136 , it was infact good to go .

I will work on this pull-request as you suggested , it will be purely for registration of plugins

PR #150, opened 2019-02-14
fix missing add of FaceliftEnum.h to FaceliftModelLib

derived-coder:
When using a cmake generated installation of facelift
this header was not installed properly.

Remark: A unit test for testing the facelift installation would be nice.


jacky309:
Thanks, but that is fixed already

derived-coder:
Okay, wasn't aware of this. However, as this long lines of header and sources, it better to align the list vertically instead of horizontally. Harder to maintain so, IMHO.

jacky309:
I agree that your formatting is better. Could you please rebase your commit ?

PR #156, opened 2019-02-19
Fixes

jacky309:


src/model/FaceliftImplementationBase.h
```diff
@@ -0,0 +1,57 @@
...
+#endif
+
+
+namespace facelift {
+
+struct PropertyInit {
+
+    typedef void (QObject::*ChangeSignal)();
+
+    facelift::PropertyBase &prop;
```
derived-coder: inconsistency between pointer an reference placement:
`facelift::PropertyBase &prop;`
` const char* name;`

src/model/FaceliftImplementationBase.h
```diff
@@ -0,0 +1,57 @@
...
+#else
+#  define FaceliftModelLib_EXPORT Q_DECL_IMPORT
+#endif
+
+
+namespace facelift {
+
+struct PropertyInit {
+
+    typedef void (QObject::*ChangeSignal)();
```
derived-coder: use `using`, no typedef anymore
using ChangeSignal = void (QObject::*)();

src/model/FaceliftImplementationBase.h
jacky309: Feel free to provide an alternative

Issue #166, opened 2019-03-13
Build broken with gcc 7.3


vifactor:
src/model/FaceliftModel.h:349:18: error: ‘function’ in namespace ‘std’ does not name a template type

Requires #include <functional> in the header

PR #167, opened 2019-03-13
Minor fixes to examples, Facelift model and code generation script

vifactor:
due to missing header (Issue #166)

### DIFF
--- a/src/model/FaceliftModel.h
+++ b/src/model/FaceliftModel.h
@@ -31,6 +31,8 @@
 #pragma once
 
 #include <memory>
+#include <functional>
+
 #include <QObject>
 #include <QDebug>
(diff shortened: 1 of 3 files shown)


src/model/FaceliftModel.h
```diff
@@ -39,7 +39,6 @@
 #include <qqml.h>
 #include <QPointer>
 
-
```
weimerb: No substantial change, pleas remove

vifactor:
Done.

src/model/FaceliftModel.h
vifactor: I squashed commits but for some reason CI queued a job but the status has been pending for some time. Should I do something about that?

PR #168, opened 2019-03-14
Fix MyPackage example

vifactor:
Misprint in method call


vifactor:
Comes from another branch

PR #171, opened 2019-03-27
ServiceWrapper: emit readyChanged each time when impl has been changed

SamaelD:
ready() property might be changed in runtime due to some conditions, but
ServiceWrapper doesn't reflect those changes, it only emits readyChanged
singal only once when wrapped object was changed. And dependency on
ready property is not working

### DIFF
--- a/src/model/ServiceWrapper.h
+++ b/src/model/ServiceWrapper.h
@@ -83,6 +83,8 @@ protected:
             WrappedType *previouslyWrapped = m_wrapped;
             __clearConnections__();
             m_wrapped = wrapped;
+            addConnection(QObject::connect(this->wrapped(), &WrappedType::readyChanged,
+                          [this] { emit this->readyChanged(); }));
             facelift::ServiceWrapperBase::setWrapped(*this, m_wrapped);
             bind(wrapped, previouslyWrapped);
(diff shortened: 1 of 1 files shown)


SamaelD:
> @SamaelD: Will you submit a pull request?

Yes, submitted

src/model/ServiceWrapper.h
```diff
@@ -83,6 +83,7 @@ class ServiceWrapper : public WrappedType, public ServiceWrapperBase
             WrappedType *previouslyWrapped = m_wrapped;
             __clearConnections__();
             m_wrapped = wrapped;
+            addConnection(QObject::connect(this->wrapped(), &WrappedType::readyChanged, [this] { emit this->readyChanged(); }));
```
weimerb: Style: limit lines to 120 chars

SamaelD:
> Commit message nitpicks: we usually start with a verb (e.g. "Emit readyChanged in ServiceWrapper when implementation changed"); "only" redundant, last sentence not quite clear and starts with "And".

Done

src/model/ServiceWrapper.h
SamaelD: Done

PR #173, opened 2019-04-03
Move logging categories into facelift namespace

jacky309:
And do not use those categories from examples


examples/AddressBook/models/cpp/AddressBookCpp.h
```diff
@@ -58,7 +58,7 @@ class AddressBookImplementation : public AddressBook
 
     void createNewContact() override
     {
-        qCDebug(LogGeneral) << "C++ createNewContact called";
```
rakeshsadhu: why categorized logs are removed from examples?

jacky309: Facelift categories are not meant to be used by application code. Our examples should be written accordingly.

PR #178, opened 2019-04-25
Add method to register an uncreatable type

weimerb:


tests/combined/tst_combined_local.qml
```diff
@@ -81,4 +81,12 @@ TestCase {
     function test_signals() {
         Check.signals();
     }
+
+    property CombinedInterfaceUncreatable combiUncreatable
```
jacky309: We don't want a type called "XXXUncreatable"

doc/page-reference.h
```diff
@@ -166,8 +166,16 @@ the QML type \c MyInterfaceImplementationBase):
...
-engine. This is done implicitly without mentioning this type.
+
+If only the type itself is needed for instance to define a typed property or only enum values from
+the type are needed, the type can be instantiated as "uncreatable":
+\code
+facelift::registerUncreatableQmlComponent<MyInterface>(uri, "MyInterfaceUncreatable");
+\endcode
+
+Note again that the above calls actually register a \c MyInterfaceQMLFrontend (derived) type to the
+QML engine internally. This type exactly reflects all the properties, methods and signals defind in
```
jacky309: defind

doc/page-reference.h
```diff
@@ -166,8 +166,16 @@ the QML type \c MyInterfaceImplementationBase):
...
-Note again that both calls actually register a \c MyInterfaceQMLFrontend derived type to the QML
-engine. This is done implicitly without mentioning this type.
+
+If only the type itself is needed for instance to define a typed property or only enum values from
+the type are needed, the type can be instantiated as "uncreatable":
+\code
+facelift::registerUncreatableQmlComponent<MyInterface>(uri, "MyInterfaceUncreatable");
+\endcode
+
+Note again that the above calls actually register a \c MyInterfaceQMLFrontend (derived) type to the
```
jacky309: I can only see one call above

src/model/QMLFrontend.h
```diff
@@ -183,6 +185,20 @@ QObject *singletonGetterByFunction(QQmlEngine *engine, QJSEngine *scriptEngine)
 }
 
 
+/*!
+ * Register the given interface QML implementation as an ucreatable QML type.
```
jacky309: does not register a QML implementation

tests/combined/tst_combined_local.qml
weimerb: hmm, that's for the sake of testing. It makes it clear what is tested. What would you propose instead? I consider CombinedInterfaceUncreatable better than plain CombinedInterface here. The latter could be used, since the creatable type is called CombinedInterfaceAPI - which you dont like either, I know.

src/model/QMLFrontend.h
weimerb: That's what I always say. It is basically copy-pasted from the other registration functions docu, that you wrote. I always stress that what is actually registered is not the interface implementation but the frontend type. You are a bit vague here...

doc/page-reference.h
```diff
@@ -166,8 +166,16 @@ the QML type \c MyInterfaceImplementationBase):
 \code
 facelift::registerQmlComponent<MyInterfaceImplementationBaseQML>(uri, "path/to/MyInterfaceImplementation.qml", "MyInterface");
 \endcode
-Note again that both calls actually register a \c MyInterfaceQMLFrontend derived type to the QML
-engine. This is done implicitly without mentioning this type.
+
+If only the type itself is needed for instance to define a typed property or only enum values from
+the type are needed, the type can be instantiated as "uncreatable":
```
jacky309: "the type can be instantiated" ? No it can not

examples/AddressBook/AddressBookPlugin.cpp
```diff
@@ -43,11 +43,14 @@ void AddressBookPlugin::registerTypes(const char *uri)
...
     Module::registerQmlTypes(uri);
-    Module::registerUncreatableQmlTypes(uri);
+
+    // Register uncreatable type as "AddressBook" (default name)
+    facelift::registerUncreatableQmlComponent<AddressBook>(uri);
 
     // We are registering the model types here, which can be used by the UI code.
     // The decision to register a dummy, QML, or C++ implementation should be taken here
-    facelift::registerQmlComponent<AddressBookCppWithProperties>(uri, "AddressBook");
+    facelift::registerQmlComponent<AddressBookCppWithProperties>(uri, "AddressBookImpl");
```
rakeshsadhu: Didn't we agree not to register plugin name ending with *Impl ? or correct me if i am wrong

PR #197, opened 2019-06-21
Fix QML engine reference in nested QML adapters

AShakhmut:
Explicitly set QML context for nested adapters
to be able to get valid QML engine reference via
qmlEngine() global function.

### DIFF
--- a/src/model/QMLAdapter.h
+++ b/src/model/QMLAdapter.h
@@ -258,7 +258,7 @@ typename ProviderType::QMLAdapterType *getQMLFrontend(ProviderType *provider)
 }
 
 template<typename ProviderType>
-typename ProviderType::QMLAdapterType *getQMLAdapter(ProviderType *provider)
+typename ProviderType::QMLAdapterType *getQMLAdapter(ProviderType *provider, QQmlEngine* engine = nullptr)
 {
     if (provider == nullptr) {
         return nullptr;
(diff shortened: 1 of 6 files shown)


jacky309:
Please delete one of these duplicate pull requests

tests/combined/plugin/CombinedTestsPlugin.cpp
```diff
@@ -59,8 +59,8 @@ void CombinedTestsPlugin::registerTypes(const char *uri)
             "/impl/qml/CombinedTestsQmlImplementation.qml", "CombinedInterfaceSingleton");
 #else
     facelift::registerQmlComponent<CombinedInterfaceImplementation>(uri, "CombinedInterfaceAPI");
-
     facelift::registerSingletonQmlComponent<CombinedInterfaceImplementation>(uri, "CombinedInterfaceSingleton");
+    facelift::registerUncreatableQmlComponent<OtherInterfaceImplementation>(uri);
```
jacky309: is that really needed ?

AShakhmut: no, I'll remove

src/model/QMLAdapter.h
```diff
@@ -258,7 +258,7 @@ typename ProviderType::QMLAdapterType *getQMLFrontend(ProviderType *provider)
 }
 
 template<typename ProviderType>
-typename ProviderType::QMLAdapterType *getQMLAdapter(ProviderType *provider)
+typename ProviderType::QMLAdapterType *getQMLAdapter(ProviderType *provider, QQmlEngine* engine = nullptr)
```
jacky309: A better approach is that the QQmlEngine instance used by a sub QMLAdapter is obtained from the parent's QMLAdapter. This is implemented in https://github.com/Pelagicore/facelift/pull/200

PR #204, opened 2019-08-16
Extended documentation

bitmouse:
Added documentation extension to some doxygen pages.
Some tests were performed to verify current behaviours
which were the base for this extension.

### DIFF
--- a/doc/page-annotations.h
+++ b/doc/page-annotations.h
@@ -26,39 +26,39 @@ definition files. An annotation can be used to enable a feature that is disabled
 
 The following annotations can be used in interface definitions:
 
-|Annotation                     |              Description                                    |
-|------------------------------ |-------------------------------------------------------------|
-|\@ipc-sync: true               | Enables the creation of a synchronous/blocking IPC proxy |
-|\@ipc-async: true              | Enables the creation of a asynchronous/non-blocking IPC proxy |
-|\@qml-implementation: true     | Enables the creation and registration of a QML component which can be used to write an implementation of the interface using QML/JS languages |
+|Annotation                     |              Description                                    | Supported over IPC |
+|------------------------------ |-------------------------------------------------------------|--------------------|
+|\@ipc-sync: true               | Enables the creation of a synchronous/blocking IPC proxy    | Yes                |
+|\@ipc-async: true              | Enables the creation of a asynchronous/non-blocking IPC proxy | Yes              |
+|\@qml-implementation: true     | Enables the creation and registration of a QML component which can be used to write an implementation of the interface using QML/JS languages | Yes |
 
 
 \section Structure_annotations Structure
 
 The following annotations can be used in structure definitions:
 
-|Annotation                     |              Description                                    |
-|-------------------------------|-------------------------------------------------------------|
-|\@serializable: true           | Adds serialization/deserialization capability to the structure |
-|\@qml-component: true          | Enables the creation and registration of a creatable QML component for the corresponding structure |
+|Annotation                     |              Description                                    | Supported over IPC |
+|-------------------------------|-------------------------------------------------------------|--------------------|
+|\@serializable: true           | Adds serialization/deserialization capability to the structure | Yes             |
+|\@qml-component: true          | Enables the creation and registration of a creatable QML component for the corresponding structure | Yes |
 
 
 \section Property_annotations Property
 
 The following annotations can be used in property definitions:
 
-|Annotation                     |              Description                                    |
-|-------------------------------|-------------------------------------------------------------|
-|\@hasReadyFlag: true           | See \ref property-ready-flag                                |
+|Annotation                     |              Description                                    | Supported over IPC |
+|-------------------------------|-------------------------------------------------------------|--------------------|
+|\@hasReadyFlag: true           | See \ref property-ready-flag                                | No                 |
 
 
 \section Method_annotations Method
 
 The following annotations can be used in method definitions:
 
-|Annotation                     |              Description                                    |
-|-------------------------------|-------------------------------------------------------------|
-|\@async: true                  | See \ref async-methods                                      |
+|Annotation                     |              Description                                    | Supported over IPC |
+|-------------------------------|-------------------------------------------------------------|--------------------|
+|\@async: true                  | See \ref async-methods                                      | Yes                |
 
 
 \section Example_annotations Example
@@ -84,5 +84,8 @@ struct MyStruct
     string s;
 }
 \endcode
+
+\section Annotations_interactions Interactions
+There are no interactions between annotations.
 
 */
--- a/doc/page-features.h
+++ b/doc/page-features.h
@@ -52,6 +52,10 @@ lambda function:
     }
 \endcode
 
+\subsection async-sub4 Async with synchronous IPC proxy
+"@ipc-sync: true" generates a proxy which has exactly the same API as the interface on the server side. \n
+"@ipc-async: true" generates a proxy whose API is like if all methods were marked @async.
+
 \section property-ready-flag Ready Flag
 
 \subsection sub1 QFace Syntax
@@ -103,7 +107,6 @@ IReadyFrontend {    // or whatever name the frontend is registered with
 Just to be clear: if you have several properties annotated with hasReadyFlag in your interface,
 you will not get indivdual signals for the ready flag of each of the properties, but get a
 \c readyFlagsChanged signal whenever the ready flag of a single property changes.
-
 
 \section userData User Data in Structures
 
--- a/doc/page-cmakeoptions.h
+++ b/doc/page-cmakeoptions.h
@@ -1,0 +1,33 @@
+/**********************************************************************
+**
+** Copyright (C) Luxoft Sweden AB 2019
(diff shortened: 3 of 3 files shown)


doc/page-annotations.h
```diff
@@ -85,4 +85,31 @@ struct MyStruct
...
 \endcode
 
+\section Annotations_impact Annotations impact
+
+Annotations can have different impact on IPC vs local methods of communications or C++ vs QML code.
+There is simplified description of this impact in table below. For detailed reference see above.
+
+|Annotation           | IPC | local | comments                 | C++ | QML| comments |
+|---------------------|-----|-------|--------------------------|-----|----|----------|
+|\@ipc-sync           |  +  | N/A   | generates ipc sync proxy | +/- | +  | method can be implemented in c++ but direct C++ usage requires extra effort |
```
jacky309: I do not understand what "method can be implemented in c++ but direct C++ usage requires extra effort" means.
Please use the existing tables describing annotations and add whatever information you believe is missing.
Annotations apply to different kind of object (interfaces, structs) and should not be mixed together in one single table.

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
 \param URI The URI of the Target i.e qml plugin
 \param VERSION The target i.e Qml plugin version, if not provided it defaults to 1.0
 
+\section cmake-params CMake options
+
+| Option                            | description                                                         |
+|-----------------------------------|---------------------------------------------------------------------|
+| FACELIFT_DISABLE_DBUS_IPC         | Disables IPC during build. Cannot use with FACELIFT_ENABLE_DBUS_IPC |
+| FACELIFT_ENABLE_CODEGEN           | generates code; should be used in rare cases only e.g. cross compilation of already generated code |
+| FACELIFT_ENABLE_DBUS_IPC          | Enables IPC during build. Cannot use with FACELIFT_DISABLE_DBUS_IPC |
```
jacky309: this enables DBUS IPC and this is not only about build time.

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
 \param URI The URI of the Target i.e qml plugin
 \param VERSION The target i.e Qml plugin version, if not provided it defaults to 1.0
 
+\section cmake-params CMake options
+
+| Option                            | description                                                         |
+|-----------------------------------|---------------------------------------------------------------------|
+| FACELIFT_DISABLE_DBUS_IPC         | Disables IPC during build. Cannot use with FACELIFT_ENABLE_DBUS_IPC |
+| FACELIFT_ENABLE_CODEGEN           | generates code; should be used in rare cases only e.g. cross compilation of already generated code |
```
jacky309: This option enables the build and installation of the code generator. it does not "generate code". It can be useful to  set it to OFF when cross-compiling. mentioning "should be used in rare cases" has not value here. Just describe when it can be useful.

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
...
+| FACELIFT_ENABLE_DBUS_IPC          | Enables IPC during build. Cannot use with FACELIFT_DISABLE_DBUS_IPC |
+| FACELIFT_BUILD_EXAMPLES           | builds examples attached to project                                 |
+| FACELIFT_BUILD_TESTS              | builds and run tests                                                |
```
jacky309: it does not run tests

doc/page-annotations.h
```diff
@@ -85,4 +85,11 @@ struct MyStruct
 }
 \endcode
 
+\section Annotations_interactions Interactions
+All annotations currently should have no interactions between each other.
```
jacky309: the use of "should" suggests that you are not 100% sure. This is confirmed by the statement below.
The documentation should be clear.

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
 \param URI The URI of the Target i.e qml plugin
 \param VERSION The target i.e Qml plugin version, if not provided it defaults to 1.0
 
+\section cmake-params CMake options
```
jacky309: This page describes the CMake functions/macros which Facelift provides to any package.
The configure options do not really belong here. I would rather have them in a separate "configuration options" section

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
...
+
+| Option                            | description                                                         |
+|-----------------------------------|---------------------------------------------------------------------|
+| FACELIFT_DISABLE_DBUS_IPC         | Disables IPC. The IPC part will not be built. Cannot use with FACELIFT_ENABLE_DBUS_IPC |
```
jacky309: This option disables the usage of DBUS. Some IPC classes are still available but no communication over DBUS is possible, which means no communication with another process is possible.

doc/page-cmakefunctions.h
```diff
@@ -122,4 +122,17 @@ This function adds a test with the name \e TargetName.The parameter descriptions
...
+| FACELIFT_ENABLE_CODEGEN           | Enables code generation. Should be disabled in rare cases, e.g. may be helpful during cross-compiling. |
+| FACELIFT_ENABLE_DBUS_IPC          | Enables IPC. The IPC part will be built. Cannot use with FACELIFT_DISABLE_DBUS_IPC |
```
jacky309: see my comment above

doc/page-features.h
```diff
@@ -104,6 +109,65 @@ Just to be clear: if you have several properties annotated with hasReadyFlag in
 you will not get indivdual signals for the ready flag of each of the properties, but get a
 \c readyFlagsChanged signal whenever the ready flag of a single property changes.
 
+\subsection sub5 Current behaviour
```
jacky309: Let's not document a behavior which we consider as broken.

doc/page-features.h
```diff
@@ -52,6 +52,11 @@ lambda function:
     }
 \endcode
 
+\subsection async-sub3 Async in synchronous interface
```
jacky309: There is nothing called "synchronous interface"

doc/page-features.h
```diff
@@ -52,6 +52,11 @@ lambda function:
...
+Asynchronous flag can be used with synchronous interface which changes behaviour of annotated method to be alwyas asynchronous.
```
jacky309: "@ipc-sync: true" generates a proxy which has exactly the same API as the interface on the server side.
"@ipc-async: true" generates a proxy whose API is like if all methods were marked @async.

doc/page-cmakefunctions.h
bitmouse: So the statement is true then, there is no IPC. I will clarify more.

doc/page-features.h
bitmouse: ok

doc/page-features.h
```diff
@@ -52,6 +52,13 @@ lambda function:
     }
 \endcode
 
+\subsection async-sub4 Async with synchronous IPC proxy
+Asynchronous flag can be used with synchronous proxy, which changes behaviour of annotated method to be alwyas asynchronous.
```
jacky309: "Asynchronous flag" ? which one ? These sentences are confusing. The last 2 lines of this paragraph tell everything. Nothing else is needed, IMO .

doc/page-annotations.h
```diff
@@ -85,4 +85,11 @@ struct MyStruct
 }
 \endcode
 
+\section Annotations_interactions Interactions
+All annotations currently have no interactions between each other with one exception described below.
+
+If IPC synchronous proxy is generated and method has \@async: true flag then this method will always be generated as asynchronous. It does not depend to proxy type.
```
jacky309: confusing. See my comment below.

doc/page-annotations.h
```diff
@@ -26,39 +26,39 @@ definition files. An annotation can be used to enable a feature that is disabled
...
-|------------------------------ |-------------------------------------------------------------|
-|\@ipc-sync: true               | Enables the creation of a synchronous/blocking IPC proxy |
-|\@ipc-async: true              | Enables the creation of a asynchronous/non-blocking IPC proxy |
-|\@qml-implementation: true     | Enables the creation and registration of a QML component which can be used to write an implementation of the interface using QML/JS languages |
+|Annotation                     |              Description                                    | Supported over IPC |
+|------------------------------ |-------------------------------------------------------------|--------------------|
+|\@ipc-sync: true               | Enables the creation of a synchronous/blocking IPC proxy    | Yes                |
+|\@ipc-async: true              | Enables the creation of a asynchronous/non-blocking IPC proxy | Yes              |
+|\@qml-implementation: true     | Enables the creation and registration of a QML component which can be used to write an implementation of the interface using QML/JS languages | N/A |
```
jacky309: IPC : Yes

PR #224, opened 2019-11-22
Fixes

jacky309:


src/ipc/ipc-common/LocalProviderBinder.h
```diff
@@ -66,9 +66,9 @@ class LocalProviderBinder : public QObject
     void onLocalAdapterUnavailable(QString objectPath, NewIPCServiceAdapterBase *adapter)
     {
         Q_UNUSED(objectPath);
-        auto service = m_interfaceManager.serviceMatches(m_proxy.objectPath(), adapter);
-        if (service) {
+        if (m_adapter == adapter) {
```
bitmouse: Please add comment telling that you compare pointers, not values. Now it is clear, but after some time may look like error.

src/model/FaceliftConversion.h
```diff
@@ -470,7 +470,7 @@ struct TypeHandler<QList<ElementType> >
         Q_UNUSED(receiver);
         Q_UNUSED(function);
         Q_UNUSED(connections);
-        static_assert(sizeof(ElementType) >= 0, "Missing specialization");
+        static_assert(sizeof(ElementType) < 0, "Missing specialization");
```
bitmouse: This looks suspicious. Could you explain this change?

jacky309: This fixes a mistake. sizeof() always return a positive value so the assertion was always true, which means it was useless.

PR #236, opened 2020-01-23
Change assignment from string to enum

bitmouse:
Make comparison case insensitive


bitmouse:
Not a valid solution. EOT.

PR #240, opened 2020-02-04
Revert "Use local IPC adapter instead of DBUS one if possible"

bitmouse:
Reverts Pelagicore/facelift#239


bitmouse:
Not needed anymore.

PR #248, opened 2020-03-02
Fixes

jacky309:


src/model/FaceliftProperty.h
```diff
@@ -51,6 +51,9 @@ class FaceliftModelLib_EXPORT PropertyBase
 
     virtual ~PropertyBase();
 
+    PropertyBase(PropertyBase const &) = delete;
```
bitmouse: Could you also take care of &&? (rule of five)

jacky309: done

PR #251, opened 2020-03-05
Improve IPC logging

jacky309:


src/ipc/dbus/DBusIPCProxy.h
```diff
@@ -121,6 +121,7 @@ class DBusIPCProxy : public IPCProxyBase<InterfaceType>, protected DBusRequestHa
             this->readyChanged();
         }
 
+        qCDebug(LogIpc) << "DBus service registered" << isRegistered << this;
```
bitmouse: Would be beneficial to print also the service name

PR #259, opened 2020-04-24
Disable IPC proxies when not used

bitmouse:


src/ipc/dbus/DBusIPCProxyBinder.cpp
```diff
@@ -171,6 +171,14 @@ void DBusIPCProxyBinder::checkRegistry()
     }
 }
 
+bool DBusIPCProxyBinder::isAllowedToConnect()
```
jacky309, line 174: "Allowed" suggests that there is some kind of permission somewhere, which is not the case.
I don't see the need to define a new method for this expression which is used only once anyway.

src/ipc/dbus/DBusIPCProxyBinder.cpp
```diff
@@ -145,29 +145,31 @@ void DBusIPCProxyBinder::onServiceNameKnown()
...
 {
-    auto registryObjects = m_registry.objects(isSynchronous());
-    if (registryObjects.contains(objectPath())) {
-        auto serviceName = registryObjects[objectPath()];
-        if (serviceName != m_serviceName) {
-            m_serviceName = serviceName;
-
-            if (!m_serviceName.isEmpty() && !m_interfaceName.isEmpty() && manager().isDBusConnected()) {
-                onServiceNameKnown();
+    if (enabled()){
```
jacky309, line 148: formatting

PR #268, opened 2020-05-08
Don't connect proxy if service is available

bitmouse:


src/ipc/ipc-common/IPCProxy.h
```diff
@@ -64,15 +64,21 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
                     for (auto& proxy : m_ipcProxies) {
                         auto proxyAdapterIPCBinder = proxy.ipcBinder;
                         if (proxyAdapterIPCBinder != nullptr) {
+                            proxyAdapterIPCBinder->setEnabled(false); // prevent connecting to server in next two steps
                             proxyAdapterIPCBinder->setObjectPath(objPath);
                             if(!serviceName.isEmpty()) {
                                 proxyAdapterIPCBinder->setServiceName(serviceName);
                             }
+                            proxyAdapterIPCBinder->setEnabled(!serviceFound);
                             proxyAdapterIPCBinder->connectToServer();
```
jacky309, line 77: calling "connectToServer()" on a proxy which previously had a "setEnabled(false)" is very weird.

src/ipc/ipc-common/IPCProxyBinderBase.h
```diff
@@ -56,6 +56,10 @@ class FaceliftIPCCommonLib_EXPORT IPCProxyBinderBase : public QObject
         return m_serviceName;
     }
 
+    /**
+     * @brief setServiceName sets the service name, then tries to connect
```
jacky309, line 60: the meaning of "try to connect" is vey unclear

src/ipc/ipc-common/IPCProxy.h
bitmouse: Agree. However it is already called twice at this place. Anyway, I will add 'if'.

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -64,15 +64,23 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
                 if (m_localProviderBinder.provider() == nullptr) {
                     auto serviceName = ipc()->serviceName();
                     auto objPath = ipc()->objectPath();
+                    bool serviceFound = false;
                     for (auto& proxy : m_ipcProxies) {
                         auto proxyAdapterIPCBinder = proxy.ipcBinder;
                         if (proxyAdapterIPCBinder != nullptr) {
+                            proxyAdapterIPCBinder->setEnabled(false); // prevent connecting to server in next two steps
```
jacky309, line 71: The logic which handles the enabling/disabling of sub-proxies is already implemented in "refreshProvider()". Please move your code their

bitmouse: This would require calling refreshProvider first on unconnected proxies or calling connectToServer in refreshProvider. Current decoupling of concerns is weak but this change would make it even weaker. Moreover hidden connectToServer calls are present in multiple methods, which makes it even less predictable and harder to read in case the code is moved there. 
This class requires refactoring which will be covered by different commits, I disagree to do it in this commit. It just makes no sense.

jacky309: No refactoring is needed : https://github.com/Pelagicore/facelift/pull/274/commits/d33eb8873a820fdf6565253be4d1242daca52efc

PR #274, opened 2020-05-24
Fixes

jacky309:


src/model/PropertyVerifier.h
```diff
@@ -57,13 +57,13 @@ void checkProperty(OwnerType* object, Getter getter, Signal signal, const char*
 template<typename OwnerType, typename PropertyType>
 static void checkProperty(const PropertyInterface<OwnerType, PropertyType>& property, const char* propertyName)
 {
-    checkProperty(property.object(), property.getter(), property.signal(), propertyName);
+    facelift::checkProperty(property.object(), property.getter(), property.signal(), propertyName);
```
bitmouse, line 60: we should reduce explicit scoping, not add it. Please remove facelift:: as this is not needed. It only creates mess in the code. Same below. If it causes troubles, use "using" directive.

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,45 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
+                        QObject::connect(ipcBinder, &IPCProxyBinderBase::serviceAvailableChanged, this, &IPCProxy::refreshProvider);
                     }
                 }
                 this->refreshProvider();
-            });
-
+        });
     }
 
     void refreshProvider() override
```
bitmouse, line 74: this function breaks SRP now https://en.wikipedia.org/wiki/Single-responsibility_principle
Please decouple responsibilities.
This function is also getting too big/too long. Please split into smaller pieces.

src/ipc/ipc-common/Registry.h
```diff
@@ -33,39 +33,61 @@
 #include <QMap>
 #include <functional>
 #include <QVector>
+#include <QDebug>
+
+struct RegistryEntry {
```
bitmouse, line 38: this should be called Listener

src/ipc/ipc-common/Registry.h
```diff
@@ -33,39 +33,61 @@
...
+    using ListenerFunction = std::function<void()>;
```
bitmouse, line 39: do we really need this "using"? It makes things less readable. 
"Function" is enough as a name.

src/ipc/ipc-common/Registry.h
```diff
@@ -33,39 +33,61 @@
...
-            if (p) {
-                auto *receiver = p.data();
-                (receiver->*receiverMethod)();
+        RegistryEntry e;
+        e.context = receiver;
+        e.function = [receiverMethod, receiver]() {
+            (receiver->*receiverMethod)();
+        };
+        listeners.append(e);
+        QObject::connect(receiver, &QObject::destroyed, [this, receiver,key] () {
```
bitmouse, line 60: Potential source of crash if registry is destroyed before connection (unlikely, but not excluded). Please create private slot for this, it will be more elegant than lambda. It will also follow SRP then. 
One more reason to extract this lambda: having method removeListener will improve API.

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,45 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
+        auto serviceName = ipc()->serviceName();
+        auto objPath = ipc()->objectPath();
+
+        for (auto& proxy : m_ipcProxies) {
+            auto ipcBinder = proxy.ipcBinder;
+            if (ipcBinder != nullptr) {
+                ipcBinder->setEnabled(!disableFollowingProxies);
+
+                if (!disableFollowingProxies) {
+                    ipcBinder->setObjectPath(objPath);
```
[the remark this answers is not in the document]
jacky309: connectToServer() is not called internally in that case. You might be referring to an old behavior.

src/ipc/ipc-common/Registry.h
jacky309: Added "owner" argument.
"key" needs to be captured to avoid iteration over the whole map.

src/ipc/ipc-common/Registry.h
jacky309: Removed

src/ipc/ipc-common/Registry.h
jacky309: Done

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,45 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
-            });
-
+        });
     }
 
     void refreshProvider() override
     {
         Q_ASSERT(m_defaultProvider);
         InterfaceType *provider = m_defaultProvider;
+        bool disableFollowingProxies = false;
```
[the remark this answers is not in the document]
jacky309: done

src/model/PropertyVerifier.h
jacky309: not providing any namespace caused an ambiguous reference error.

src/ipc/ipc-common/IPCProxy.h
jacky309: Introduced sub-method

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,48 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
-                    break;
+            enableFollowingProxies = false;
+        }
+
+        for (auto& proxy : m_ipcProxies) {
+            auto ipcBinder = proxy.ipcBinder;
+            if (ipcBinder != nullptr) {
+                ipcBinder->setEnabled(enableFollowingProxies);
+
+                if (enableFollowingProxies) {
```
bitmouse, line 99: It won't call setObjectPath on every ipcBinder

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,48 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
+            ipcBinder->setServiceName(serviceName);
+        }
+        ipcBinder->connectToServer();
     }
 
     void refreshProvider() override
     {
         Q_ASSERT(m_defaultProvider);
         InterfaceType *provider = m_defaultProvider;
+        bool enableFollowingProxies = true;
```
bitmouse, line 88: serviceFound is nicer name IMO (suggestion only)

src/ipc/ipc-common/Registry.h
```diff
@@ -29,43 +29,68 @@
...
 #pragma once
 
+#include <QObject>
 #include <QString>
 #include <QMap>
 #include <functional>
 #include <QVector>
+#include <QDebug>
+
+struct RegistryListenerEntry {
```
bitmouse, line 39: just Listener. Less means more. If you afraid names conflict, please namespace it.

src/ipc/ipc-common/Registry.h
```diff
@@ -112,17 +137,13 @@ class Registry {
...
         }
 
     }
 
 private:
     QMap<Key, Type> m_content;
-    mutable QMap<Key, QVector<ListenerFunction>> m_listeners;  // "mutable" since adding a listener is not considered as a change
-
+    mutable QMap<Key, QVector<RegistryListenerEntry>> m_listeners;  // "mutable" since adding a listener is not considered as a change
+    QObject* m_owner = nullptr;
```
bitmouse, line 148: No need to initialize it here with nullptr if you init it in the constructor (there is only one).

src/ipc/ipc-common/IPCProxy.h
jacky309: you are right, and I seen nothing wrong here.

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,45 +61,48 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
                     }
                 }
                 this->refreshProvider();
-            });
+        });
+    }
 
+    void enableBinder(IPCProxyBinderBase * ipcBinder)
+    {
+        ipcBinder->setObjectPath(ipc()->objectPath());
```
bitmouse, line 76: It will call connectToServer succesfully if isReadyToConnect returns true. Then our connectToServer is not needed below. IMO this hidden (and early) connection should be somehow disabled.

src/ipc/ipc-common/Registry.h
bitmouse: Bump

src/ipc/ipc-common/Registry.h
bitmouse: Bump.

src/ipc/ipc-common/IPCProxy.h
bitmouse: Alternatively add relevant comment here

src/ipc/ipc-common/Registry.h
jacky309: Class turned into an inner class of Registry

src/ipc/ipc-common/Registry.h
jacky309: Not needed with the current code indeed, but it is a precaution to avoid leaving that pointer uninitialized if the constructor happens to be changed later.

src/ipc/ipc-common/IPCProxy.h
```diff
@@ -61,46 +61,54 @@ class IPCProxy : public WrapperType, public IPCProxyNewBase
...
                     }
                 }
                 this->refreshProvider();
-            });
+        });
+    }
 
+    void enableBinder(IPCProxyBinderBase * ipcBinder)
+    {
+        if (!ipcBinder->enabled()) {
```
bitmouse, line 76: I think this will cause problem with default value in IPCProxyBinderBase "bool m_enabled = true;"

PR #276, opened 2020-06-03
Initialization of variables

pfagner:
Add initialization of member variables

### DIFF
--- a/src/ipc/local/LocalIPCProxy.h
+++ b/src/ipc/local/LocalIPCProxy.h
@@ -128,7 +128,7 @@ private:
     QString m_interfaceName;
     LocalIPCRequestHandler *m_serviceObject = nullptr;
     QPointer<LocalIPCServiceAdapterBase> m_serviceAdapter;
-    QMetaObject::Connection m_signalConnection;
+    QMetaObject::Connection m_signalConnection{};
     bool m_explicitServiceName = false;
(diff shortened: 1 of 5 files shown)


src/ipc/local/LocalIPCProxy.h
```diff
@@ -128,7 +128,7 @@ class FaceliftIPCLocalLib_EXPORT LocalIPCProxyBinder : public IPCProxyBinderBase
     QString m_interfaceName;
```
jacky309, line 128: why not here as well ?

src/ipc/local/LocalIPCProxy.h
```diff
@@ -128,7 +128,7 @@ class FaceliftIPCLocalLib_EXPORT LocalIPCProxyBinder : public IPCProxyBinderBase
...
     LocalIPCRequestHandler *m_serviceObject = nullptr;
     QPointer<LocalIPCServiceAdapterBase> m_serviceAdapter;
```
jacky309, line 130: and here ?

pfagner:
Actually I not sure that we have default initialization in current form and therefore we have risk that they remain uninitialized and we get UB.

bitmouse:
Is it still valid? If yes, please resolve conflicts and update PR.

pfagner:
This commit is irrelevant.

PR #277, opened 2020-06-03
Change the initialization order

pfagner:
Change the initialization of the base class

### DIFF
--- a/src/ipc/local/LocalIPCProxy.h
+++ b/src/ipc/local/LocalIPCProxy.h
@@ -160,7 +160,7 @@ public:
     template<typename Type>
     using IPCProxyType = typename Type::IPCLocalProxyType;
 
-    LocalIPCProxy(QObject *parent = nullptr) : IPCProxyBase<InterfaceType>(parent), LocalIPCProxyBase(m_ipcBinder), m_ipcBinder(*this)
+    LocalIPCProxy(QObject *parent = nullptr) : IPCProxyBase<InterfaceType>(parent), LocalIPCProxyBase(*this), m_ipcBinder(*this)
     {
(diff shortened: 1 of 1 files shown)


src/ipc/local/LocalIPCProxy.h
```diff
@@ -160,7 +160,7 @@ class LocalIPCProxy : public IPCProxyBase<InterfaceType>, protected LocalIPCProx
     template<typename Type>
     using IPCProxyType = typename Type::IPCLocalProxyType;
 
-    LocalIPCProxy(QObject *parent = nullptr) : IPCProxyBase<InterfaceType>(parent), LocalIPCProxyBase(m_ipcBinder), m_ipcBinder(*this)
+    LocalIPCProxy(QObject *parent = nullptr) : IPCProxyBase<InterfaceType>(parent), LocalIPCProxyBase(*this), m_ipcBinder(*this)
```
jacky309, line 163: Please justify why this change is needed.

pfagner: Initialization order
link: https://en.cppreference.com/w/cpp/language/constructor
The order of member initializers in the list is irrelevant: the actual order of initialization is as follows: 
1) If the constructor is for the most-derived class, virtual bases are initialized in the order in which they appear in depth-first left-to-right traversal of the base class declarations (left-to-right refers to the appearance in base-specifier lists) 
2) Then, direct bases are initialized in left-to-right order as they appear in this class's base-specifier list 
3) Then, non-static data member are initialized in order of declaration in the class definition. 
4) Finally, the body of the constructor is executed

Point of my view: we pass uninitialized m_ipcBinder to base class LocalIPCProxyBase

Please check that simplify variant: 
http://tpcg.io/OlQw7OhV

pfagner:
Wrong change.

PR #279, opened 2020-06-03
Check copy-assignment operator

pfagner:
Check copy-assignment operator is expected to perform no action on self-assignment, and to return the lhs by reference

### DIFF
--- a/codegen/facelift/templates/Struct.template.cpp
+++ b/codegen/facelift/templates/Struct.template.cpp
@@ -114,7 +114,9 @@ const {{struct}}::FieldNames {{struct}}::FIELD_NAMES = { {
 
 {{struct.name}}& {{struct.name}}::operator=(const {{struct.name}} &right)
 {
-    copyFrom(right);
+    if (this != &right) {
+        copyFrom(right);
+    }
     return *this;
 }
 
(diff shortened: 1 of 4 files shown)


codegen/facelift/templates/Struct.template.cpp
```diff
@@ -114,7 +114,9 @@ const {{struct}}::FieldNames {{struct}}::FIELD_NAMES = { {
 
 {{struct.name}}& {{struct.name}}::operator=(const {{struct.name}} &right)
 {
-    copyFrom(right);
+    if(this != &right){
```
jacky309, line 117: formatting please

src/model/FaceliftProperty.h
```diff
@@ -216,7 +216,9 @@ class TProperty : public PropertyBase
 
     Type &operator=(const Type &right)
     {
-        setValue(right);
+        if(this != &right){
```
jacky309, line 219: this condition can not be false.

jacky309:
Squash your commits please

codegen/facelift/templates/Struct.template.cpp
```diff
@@ -114,7 +114,10 @@ const {{struct}}::FieldNames {{struct}}::FIELD_NAMES = { {
 
 {{struct.name}}& {{struct.name}}::operator=(const {{struct.name}} &right)
 {
-    copyFrom(right);
+    if(this != &right)
```
jacky309, line 117: if (this != &right) {

PR #284, opened 2020-07-02
Disable redundant transitive linking with shared libraries

bapi23:
All targets which depend on shared libraries had to be re-linked before this change which is not needed.

### DIFF
--- a/cmake/faceliftMacros.cmake
+++ b/cmake/faceliftMacros.cmake
@@ -584,6 +584,10 @@ function(facelift_add_library TARGET_NAME)
                 set(LIBRARY_TYPE INTERFACE)
             endif()
             add_library(${TARGET_NAME} ${LIBRARY_TYPE} ${ALL_SOURCES})
+            if(NOT "${LIBRARY_TYPE}" STREQUAL "INTERFACE")
+                set_target_properties(${TARGET_NAME} PROPERTIES LINK_DEPENDS_NO_SHARED true)
+            endif()
+
             target_link_libraries(${TARGET_NAME} ${__INTERFACE} ${ARGUMENT_MONOLITHIC_LINK_LIBRARIES})
             _facelift_add_target_finish(${TARGET_NAME} ${TARGET_NAME})
 
@@ -698,6 +702,7 @@ function(facelift_add_executable TARGET_NAME)
     _facelift_parse_target_arguments("" "" "" ${ARGN})
     _facelift_add_target_start(${TARGET_NAME})
     add_executable(${TARGET_NAME} ${ALL_SOURCES})
+    set_target_properties(${TARGET_NAME} PROPERTIES LINK_DEPENDS_NO_SHARED true)
 
     if (NOT ${ARGUMENT_NO_INSTALL})
         install(TARGETS ${TARGET_NAME} EXPORT ${PROJECT_NAME}Targets DESTINATION ${CMAKE_INSTALL_BINDIR})


cmake/faceliftMacros.cmake
```diff
@@ -714,6 +719,7 @@ function(facelift_add_test TEST_NAME)
     set_target_properties(${TEST_NAME} PROPERTIES
         LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/tests
         RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/tests
+        LINK_DEPENDS_NO_SHARED true
```
kunichik, line 722: We have set param to true inside "facelift_add_executable" above (line 717).

bapi23: Right, I've overlooked that one - will fix it and test it again. Thanks!

bapi23: Most probably it needs to be like that since the second set_target_properties will override the first one. We need to have it in both places. Most probably the second one should be replaced by set_property to append property instead of override all properties. I need to test new approach.

bapi23: Just tested it and lit doesn't override the first one which is a good news:)

jacky309:
Why not using CMAKE_LINK_DEPENDS_NO_SHARED in your own package instead of forcing that behavior in facelift itself ?

bapi23:
> Why not using CMAKE_LINK_DEPENDS_NO_SHARED in your own package instead of forcing that behavior in facelift itself ? What makes you believe that all users of Facelift want that behavior ?

Thanks for that comment. I fully agree with you. Looks like CMake developers already tried to make it default behaviour:
https://cmake-developers.cmake.narkive.com/dTGZ68X4/should-cmake-link-depends-no-shared-be-on-by-default. 
It needs to be reverted.

bapi23:
> Why not using CMAKE_LINK_DEPENDS_NO_SHARED in your own package instead of forcing that behavior in facelift itself ? What makes you believe that all users of Facelift want that behavior ?

I see at least two problems with the current implementation of the facelift macros:

1. You will need to recreate "facelift_add_library" macro logic in the upper layer to add set_target_property only for non-interface libraries (INTERFACE library can't have LINK_DEPENDS_NO_SHARED set).
2. Library name might differ from the one you requested. Looks like it might be not an issue because it goes only for INTERFACE and OBJECT libraries (I'm not sure about OBJECT)
set(TARGET_NAME ${TARGET_NAME}_)
or
set(IMPLEMENTATION_TARGET_NAME ${TARGET_NAME}_OBJECTS)

So my questions are:
1. What were the reasons to change the given target name inside facelift_add_library?
2. Why facelift_add_library changes given library type? (logic could be more straightforward - for example, you need always provide library type and the function will only validate if the entire configuration is correct)

If I'm wrong please correct me. If I've missed any documentation which contains answers to my questions I would be glad to see it:)

jacky309:
> > Why not using CMAKE_LINK_DEPENDS_NO_SHARED in your own package instead of forcing that behavior in facelift itself ? What makes you believe that all users of Facelift want that behavior ?
> 
> I see at least two problems with the current implementation of the facelift macros:
> 
> 1. You will need to recreate "facelift_add_library" macro logic in the upper layer to add set_target_property only for non-interface libraries (INTERFACE library can't have LINK_DEPENDS_NO_SHARED set).

AFAIU, you simply need to set CMAKE_LINK_DEPENDS_NO_SHARED once for all, before calling facelift_add_library.

> 2. Library name might differ from the one you requested. Looks like it might be not an issue because it goes only for INTERFACE and OBJECT libraries (I'm not sure about OBJECT)
>    set(TARGET_NAME ${TARGET_NAME}_)
>    or
>    set(IMPLEMENTATION_TARGET_NAME ${TARGET_NAME}_OBJECTS)
> 
> So my questions are:
> 
> 1. What were the reasons to change the given target name inside facelift_add_library?
> 2. Why facelift_add_library changes given library type? (logic could be more straightforward - for example, you need always provide library type and the function will only validate if the entire configuration is correct)
The reason is the monolithic build. See the corresponding doxygen-based documentation (added by https://github.com/Pelagicore/facelift/commit/ad90feb9501b233792e3e136a789ff35afbc7e4f#diff-4636a4bb7c60c257a6c8e053e025ce49).

> 
> If I'm wrong please correct me. If I've missed any documentation which contains answers to my questions I would be glad to see it:)

Please

PR #285, opened 2020-07-03
Revert "Disable redundant transitive linking with shared libraries"

bapi23:
This reverts commit 9b7f4561a68aeecf524882d271bfaa3026c27103.


bitmouse:
Reverting previous commit after internal discussion.

PR #286, opened 2020-07-08
Add CMake LINK_DEPENDS_NO_SHARED property as a parameter to the Facel…

bapi23:
The reason behind adding LINK_DEPENDS_NO_SHARED as a parameter:
- LINK_DEPENDS_NO_SHARED can't be a global property. It might be set only for a specific target and since facelift CMake API changes the name of the target the most correct way of adding it is to add it as a parameter.
- The second solution would be to set/unset CMAKE_LINK_DEPENDS_NO_SHARED every time facelift API is called in the client but IMHO it looks more hackish and less straightforward.

### DIFF
--- a/cmake/faceliftMacros.cmake
+++ b/cmake/faceliftMacros.cmake
@@ -483,7 +483,7 @@ endmacro()
 
 function(facelift_add_library TARGET_NAME)
 
-    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" ""
+    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" "LINK_DEPENDS_NO_SHARED"
         "HEADERS_NO_INSTALL;HEADERS_GLOB_NO_INSTALL;HEADERS_GLOB_RECURSE_NO_INSTALL;PUBLIC_HEADER_BASE_PATH;MONOLITHIC_LINK_LIBRARIES"
         ${ARGN}
     )
@@ -496,6 +496,10 @@ function(facelift_add_library TARGET_NAME)
     endif()
 
     list(APPEND ARGUMENT_LINK_LIBRARIES ${ARGUMENT_MONOLITHIC_LINK_LIBRARIES})
+
+    if (DEFINED ARGUMENT_LINK_DEPENDS_NO_SHARED)
+        set(CMAKE_LINK_DEPENDS_NO_SHARED ${ARGUMENT_LINK_DEPENDS_NO_SHARED})
+    endif()
 
     unset(__INTERFACE)
     if(ARGUMENT_INTERFACE)
@@ -695,8 +699,12 @@ endfunction()
 
 
 function(facelift_add_executable TARGET_NAME)
-    _facelift_parse_target_arguments("" "" "" ${ARGN})
+    _facelift_parse_target_arguments("" "LINK_DEPENDS_NO_SHARED" "" ${ARGN})
     _facelift_add_target_start(${TARGET_NAME})
+    if (DEFINED ARGUMENT_LINK_DEPENDS_NO_SHARED)
+        set(CMAKE_LINK_DEPENDS_NO_SHARED ${ARGUMENT_LINK_DEPENDS_NO_SHARED})
+    endif()
+
     add_executable(${TARGET_NAME} ${ALL_SOURCES})
 
     if (NOT ${ARGUMENT_NO_INSTALL})
--- a/doc/page-cmakefunctions.h
+++ b/doc/page-cmakefunctions.h
@@ -50,6 +50,7 @@ facelift_add_qml_plugin(<TargetName>
                         [HEADERS_GLOB_NO_INSTALL [Globbing expressions]...]
                         [HEADERS_GLOB_RECURSE_NO_INSTALL [Globbing expressions]...]
                         [LINK_LIBRARIES lib1 [lib2...]]
+                        [LINK_DEPENDS_NO_SHARED <ON|OFF>]
                         [NO_INSTALL])
 \endcode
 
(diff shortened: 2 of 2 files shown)


cmake/faceliftMacros.cmake
```diff
@@ -584,6 +584,9 @@ function(facelift_add_library TARGET_NAME)
                 set(LIBRARY_TYPE INTERFACE)
             endif()
             add_library(${TARGET_NAME} ${LIBRARY_TYPE} ${ALL_SOURCES})
+            if(NOT "${LIBRARY_TYPE}" STREQUAL "INTERFACE" AND DEFINED ARGUMENT_DO_NOT_RELINK_SHARED)
```
bitmouse, line 587: space after if

cmake/faceliftMacros.cmake
```diff
@@ -695,14 +698,18 @@ endfunction()
...
-    _facelift_parse_target_arguments("" "" "" ${ARGN})
+    _facelift_parse_target_arguments("" "DO_NOT_RELINK_SHARED" "" ${ARGN})
     _facelift_add_target_start(${TARGET_NAME})
     add_executable(${TARGET_NAME} ${ALL_SOURCES})
 
     if (NOT ${ARGUMENT_NO_INSTALL})
         install(TARGETS ${TARGET_NAME} EXPORT ${PROJECT_NAME}Targets DESTINATION ${CMAKE_INSTALL_BINDIR})
     endif()
 
+    if(DEFINED ARGUMENT_DO_NOT_RELINK_SHARED)
```
bitmouse, line 709: space after if

jacky309:
I do not understand why you say "The second solution would be to set/unset CMAKE_LINK_DEPENDS_NO_SHARED every time facelift API is called". CMAKE_LINK_DEPENDS_NO_SHARED can be set once for all. CMake will effectively apply it to the libraries of the correct type.

bapi23:
> I do not understand why you say "The second solution would be to set/unset CMAKE_LINK_DEPENDS_NO_SHARED every time facelift API is called". CMAKE_LINK_DEPENDS_NO_SHARED can be set once for all. CMake will effectively apply it to the libraries of the correct type.

The problem is that I do not want to apply it to all targets (yes - it would be applied not only to libraries).

cmake/faceliftMacros.cmake
bapi23: done

cmake/faceliftMacros.cmake
bapi23: done

jacky309:
DO_NOT_RELINK_SHARED is meaningless and is inconsistent with CMake's LINK_DEPENDS_NO_SHARED. Keep CMake terminology.

cmake/faceliftMacros.cmake
```diff
@@ -584,6 +584,9 @@ function(facelift_add_library TARGET_NAME)
                 set(LIBRARY_TYPE INTERFACE)
             endif()
             add_library(${TARGET_NAME} ${LIBRARY_TYPE} ${ALL_SOURCES})
```
jacky309, line 586: before calling add_library, simply do:

if(DO_NOT_RELINK_SHARED)
    set(CMAKE_LINK_DEPENDS_NO_SHARED ON)
endif()

That should be all you need.

cmake/faceliftMacros.cmake
```diff
@@ -483,7 +483,7 @@ endmacro()
 
 function(facelift_add_library TARGET_NAME)
 
-    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" ""
+    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" "DO_NOT_RELINK_SHARED"
```
jacky309, line 486: LINK_DEPENDS_NO_SHARED

cmake/faceliftMacros.cmake
bapi23: done

cmake/faceliftMacros.cmake
bapi23: done

cmake/faceliftMacros.cmake
```diff
@@ -497,6 +497,10 @@ function(facelift_add_library TARGET_NAME)
 
     list(APPEND ARGUMENT_LINK_LIBRARIES ${ARGUMENT_MONOLITHIC_LINK_LIBRARIES})
 
+    if (DEFINED ARGUMENT_LINK_DEPENDS_NO_SHARED)
```
jacky309, line 500: move into _facelift_add_target_start

cmake/faceliftMacros.cmake
```diff
@@ -483,7 +483,7 @@ endmacro()
 
 function(facelift_add_library TARGET_NAME)
 
-    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" ""
+    _facelift_parse_target_arguments("SYSTEM;STATIC;SHARED;OBJECT;MODULE;MONOLITHIC_SUPPORTED;NO_MONOLITHIC" "LINK_DEPENDS_NO_SHARED"
```
jacky309, line 486: LINK_DEPENDS_NO_SHARED applies to all target types => move into _facelift_parse_target_arguments

cmake/faceliftMacros.cmake
```diff
@@ -695,8 +699,12 @@ endfunction()
 
 
 function(facelift_add_executable TARGET_NAME)
-    _facelift_parse_target_arguments("" "" "" ${ARGN})
+    _facelift_parse_target_arguments("" "LINK_DEPENDS_NO_SHARED" "" ${ARGN})
```
jacky309, line 702: undo change

cmake/faceliftMacros.cmake
```diff
@@ -695,8 +699,12 @@ endfunction()
...
     _facelift_add_target_start(${TARGET_NAME})
+    if (DEFINED ARGUMENT_LINK_DEPENDS_NO_SHARED)
```
jacky309, line 704: undo change

bitmouse:
Closing as outdated.

PR #288, opened 2020-07-16
Extend documentation

bitmouse:


doc/page-generated.h
```diff
@@ -0,0 +1,80 @@
...
+** CA 94042, USA.
+**
+** SPDX-License-Identifier: CC-BY-SA-4.0
+**
+**********************************************************************/
+
+/**
+\page page-generated Generated code structure
+
+This page refers to the code generated by Addressbook example localted in examples/AddressBook. The purpose is to provide an outline of generated code structure.
```
jacky309, line 20: localted

doc/page-generated.h
```diff
@@ -0,0 +1,80 @@
...
+
+Diagram below shows overall structure of the generated code and its relation to non-generated classes.
```
jacky309, line 22: articles please

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+**
+** SPDX-License-Identifier: CC-BY-SA-4.0
+**
+**********************************************************************/
+
+/**
+\page page-properties Properties
+
+\section properties-intro Introduction
+Properties are the part of the interface described in *.qface file. For more details about properties in *.qface files see documentation at https://github.com/Pelagicore/qface
```
jacky309, line 21: there are not only properties in qface files

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+In Facelift each property is exposed to the QML engine as Q_PROPERTY with a setter and getter functions, and a change signal.
```
jacky309, line 23: no setter for readonly

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+The property provides encapsulation for the internal data with a const getter only (except service property), which guarantees that the data will not be unnoticibly changed. Each time the setter is called, with a value that differs from the previous one, the change signal is emitted. If the previous value equals to the new value, the change signal will not be emitted.
```
jacky309, line 25: the property which you describe here is not the same as above

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+\section QML
+
+Property is exposed directly to QML engine by Q_PROPERTY macro and can be used without any additional effort.
```
jacky309, line 29: articles please !

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+**
+** SPDX-License-Identifier: CC-BY-SA-4.0
+**
+**********************************************************************/
+
+/**
+\page page-properties Properties
+
+\section properties-intro Introduction
+Properties are the part of the interface definition described in *.qface file. Eventually they will be generated as C++ and QML properties of the interface. For more details about properties in *.qface files see documentation at https://github.com/Pelagicore/qface
```
jacky309, line 21: "Interface definition files (*.qface) can define properties. See https://github.com/Pelagicore/qface for information about QFace interface definition."

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+In Facelift each property is exposed to the QML engine as Q_PROPERTY with a setter and getter functions, and a change signal. If the property is marked as 'readonly' then no setter is generated.
```
jacky309, line 23: Do not contradict yourself.

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+The facelift::Property class template provides encapsulation for the internal data with a const getter only (except for the service property), which guarantees that the data will not be unnoticibly changed. Each time the setter is called, with a value that differs from the previous one, the change signal is emitted. If the previous value equals to the new value, the change signal will not be emitted.
```
jacky309, line 25: You need to introduce the XXXImplementationBase class before talking about the facelift::Property template class.

jacky309: that line is too long

jacky309: Remove "If the previous value equals to the new value, the change signal will not be emitted.". The previous sentence is enough.

jacky309: "with a const getter only" ?? There are other methods in that class.
I doubt service properties have a specific here.

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+\section QML
+
+The property is exposed directly to the QML engine by Q_PROPERTY macro and can be used without any additional effort.
```
jacky309, line 29: are you talking about implementing an interface using QML ? Please make it clear.

doc/page-properties.h
```diff
@@ -0,0 +1,129 @@
...
+
+\section cpp C++
+
+The property in C++ is generated as one of the templated property containers: Property<T>, ListProperty<T>, ServiceProperty<T> or ModelProperty<T>.
+Additionally ListProperty and QMap<QString, T> partial template specializations add extra convenience functions: void removeAt(int i), void addElement(ElementType element) and int size() const.
```
jacky309, line 34: you already mentioned ListProperty. Why "Additionally" ??
you need to mention that QMap<QString,T> is used for map<T> properties defined in qface files.
Why are you getting into API details here ??

doc/page-properties.h
bitmouse: There is no contradiction in whole context.

doc/page-properties.h
bitmouse: > You need to introduce the XXXImplementationBase class before talking about the facelift::Property template class.

There will be separate page for it at some point. Reference will be added. Meanwhile I will add short information about it.

> that line is too long

This line is a whole paragraph in generated html, inside &lt;p&gt;&lt;/p&gt; html tags. Splitting this line would cause new paragraph generation.
> Remove "If the previous value equals to the new value, the change signal will not be emitted.". The previous sentence is enough.

OK

> "with a const getter only" ?? There are other methods in that class.

Service property returns const reference to a pointer. Sentence to be rephrased.

doc/page-properties.h
bitmouse: I am talking about property exposure to QML engine, not QML implementation. Will be rephrased.

doc/page-properties.h
bitmouse: > you already mentioned ListProperty. Why "Additionally" ??

They are methods that are not present in other specializations.

>  you need to mention that QMap<QString,T> is used for map properties defined in qface files.

OK

>  Why are you getting into API details here ??

Documentation of facelift is very poor. Code is not self explaining and is very complicated, sometimes even spaghetti code. Let's be honest - sometimes not a good quality. Developers should know what are they looking at. 
If we want to keep this project alive, the documentation must be improved. This documentation is not only for the end user but also for developers.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+\page page-properties Properties
+
+\section properties-intro Introduction
+Interface definition files (*.qface) can define properties. See https://github.com/Pelagicore/qface for information
+about QFace interface definition.
+
+In Facelift each property is exposed to the QML engine as Q_PROPERTY with a setter and getter functions, and a change
+signal. If the property is marked as 'readonly' then no setter is generated.
+
+In Facelift properties are generated as members of the &lt;Inerface name&gt;ImplementationBase C++ class.
```
jacky309, line 27: ImplementationBase have no Q_PROPERTY, but you say above that each property is exposed as Q_PROPERTY, which is confusing.
You need to introduce the role of the QMLAdapter class in order to be clear.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+The facelift::Property class template provides encapsulation for the internal data. To get the underlying value
+developer can use a getter that returns the const reference to the underlying data (in case of the service property
```
jacky309, line 29: there is nothing special to mention about service properties here. const references are used for service and "normal" types.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+it is const reference to a pointer) or one of the specialization specific access functions. The value can be changed only by
```
jacky309, line 30: the value is typically altered using the operator=, which means the property object can be used as if it was the actual underlying data.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+using a setter or one of the specialization specific access functions. This guarantees that the data will not be unnoticibly
+ changed. Each time the setter is called, with a value that differs from the previous one, the change signal is emitted.
+
+\section QML
+
+In case of C++ interface implementation, the property, genetated as a C++ code, is exposed directly to the QML engine
```
jacky309, line 36: given that we are in the "QML" section, this sentence makes no sense to me.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+The property in C++ is generated as one of the templated property containers: Property<T>, ListProperty<T>,
+ServiceProperty<T> or ModelProperty<T>. Additionally ListProperty and QMap<QString, T> partial template specializations
+add extra convenience functions: void removeAt(int i), void addElement(ElementType element) and int size() const.
+The QMap<QString, T> specialization is created for the map property defined in *.qface file.
+
+\section subservice Property ownership
+
+The service owns the property objects and maintains their life cycle.
+
+The service property returns a pointer to another service. In this case property object is not an owner of the
```
jacky309, line 50: "The service property" ?? which one ? a property does not return anything

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+
+In the example below, whenever Context object bound to the property emits signal foo(), the property value is set to
+the value returned from the lambda passed as bind() parameter.
+\code
+
+class Context : public QObject {
+  Q_OBJECT
+public:
+  Q_SIGNAL void foo();
+  facelift::Property<QString> prop;
```
jacky309, line 71: why is that a facelift::Property here ? it adds confusion.

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+specified getter function. Getter function must return underlying property type.
+
+In the example below, whenever Context object bound to the property emits signal foo(), the property value is set to
+the value returned from the lambda passed as bind() parameter.
+\code
+
+class Context : public QObject {
+  Q_OBJECT
+public:
+  Q_SIGNAL void foo();
```
jacky309, line 70: fooChanged

doc/page-properties.h
```diff
@@ -0,0 +1,145 @@
...
+  facelift::Property<QString> prop;
+};
+
+void bar(Context* ctx){
+  prop.bind([]() {
+          return QString("foo");
```
jacky309, line 76: you are returning a constant. Why do you connection a re-evaluation trigger then ? Confusing example !

doc/page-generated.h
```diff
@@ -0,0 +1,81 @@
...
+** this license, visit http://creativecommons.org/licenses/by-sa/4.0/
+** or send a letter to Creative Commons, PO Box 1866, Mountain View,
+** CA 94042, USA.
+**
+** SPDX-License-Identifier: CC-BY-SA-4.0
+**
+**********************************************************************/
+
+/**
+\page page-generated Generated code structure
```
jacky309, line 18: this documentation already exists in "page-reference.h". Add some content in there, if needed.

bitmouse: No, this content is different and its meaning is different. Also calling the other page "reference" is incorrect, it may be fixed in the future.

doc/page-properties.h
bitmouse: Will be done.

doc/page-properties.h
bitmouse: There is significant difference between const reference to value, const reference to the pointer and const reference to const pointer. In case of service property  it is writable, only pointer cannot be changed. This is the reason why it needs to be mentioned.
Also I think this should be analyzed as a possible bug, because I would expect consistent behavior, which apparently does not happen in this case.

doc/page-properties.h
bitmouse: To be added.

doc/page-properties.h
bitmouse: Section title to be changed to "QML exposure"

doc/page-properties.h
bitmouse: to be changed

doc/page-properties.h
bitmouse: it is an example property. It adds no confusion in the whole context. Developers are not morons, they can understand code.

doc/page-properties.h
```diff
@@ -0,0 +1,151 @@
...
+the value returned from the lambda passed as bind() parameter.
+\code
+
+QString getRandomQString() {...}
+
+class Context : public QObject {
+  Q_OBJECT
+public:
+  Q_SIGNAL void fooChanged();
+  facelift::Property<QString> prop;
```
jacky309, line 75: replace this unused line by the following:
QString foo() const;

doc/page-properties.h
```diff
@@ -0,0 +1,151 @@
...
+
+Typically property state change may happen on setter call. An alternative way of updating property value with a minimal
+effort is to bind it to the signal. Whenever property is notified (state change triggered), the property will call
+specified getter function. Getter function must return underlying property type.
+
+In the example below, whenever Context object bound to the property emits signal foo(), the property value is set to
+the value returned from the lambda passed as bind() parameter.
+\code
+
+QString getRandomQString() {...}
```
jacky309, line 69: delete

doc/page-properties.h
```diff
@@ -0,0 +1,151 @@
...
+class Context : public QObject {
+  Q_OBJECT
+public:
+  Q_SIGNAL void fooChanged();
+  facelift::Property<QString> prop;
+};
+
+void bar(Context* ctx){
+  prop.bind([]() {
+          auto str = getRandomQString();
```
jacky309, line 80: return ctx->foo();

doc/page-properties.h
jacky309: Remove that weird section since you already introduced the XXXQMLAdapter class in the introduction.

doc/page-properties.h
```diff
@@ -0,0 +1,151 @@
...
+with a value different than the previous one, or other operation that changes the value is called, the change signal is emitted.
+
+\section QML exposure
+
+In case of C++ interface implementation, the property, genetated as a C++ code, is exposed directly to the QML engine
+by Q_PROPERTY macro in the relevant QMLAdapter and can be used without any additional effort.
+
+\section cpp C++
+
+The property in C++ is generated as one of the templated property containers: Property<T>, ListProperty<T>,
```
jacky309, line 43: "The property " ?? which one ?

doc/page-properties.h
```diff
@@ -0,0 +1,154 @@
...
+The service property stores a pointer to another service. In this case property object is not an owner of the
+pointed service. In terms of QML engine, the service object ownership is held by C++ code.
+See https://doc.qt.io/qt-5/qqmlengine.html#ObjectOwnership-enum
+
+Other property types are owners of the underyling objects.
+
+
+\section bindings Property bindings
+
+Typically property state change may happen on setter call. An alternative way of updating property value with a minimal
```
jacky309, line 61: "property state" ??
"Typically" does not get along with "may".

doc/page-properties.h
```diff
@@ -0,0 +1,154 @@
...
+effort is to bind it to the signal. Whenever property is notified (state change triggered), the property will call
```
jacky309, line 62: The "bind" mechanism should be introduced in the same section as the other mechanism (using setValue(), =operator or other means).
Please use articles.

kunichik (approved):

PR #295, opened 2020-08-11
Add the clang-tidy is a clang-based C++ “linter” tool.

pfagner:


bitmouse (changes requested):
Some tweaking is required, e.g.
src/common/FaceliftUtils.h:67:5: warning: overloading 'operator()' is disallowed [fuchsia-overloaded-operator]

It is 100% valid for us. We are not fuchsia project and we do not follow their standard.

PR #296, opened 2020-08-18
Add sanitizers

pfagner:
- AddressSanitizer: detects addressability issues;
- MemorySanitizer: detects use of uninitialized memory;
- ThreadSanitizer: detects data races and deadlocks;
- LeakSanitizer: detects memory leaks;
- UndefinedBehaviorSanitizer: detect undefined behavior issues.

### DIFF
--- a/tests/sanitizers/test_sanitizers.sh
+++ b/tests/sanitizers/test_sanitizers.sh
@@ -1,0 +1,98 @@
+#!/bin/bash
+set -e
+# for info see 'read.me'
+
+DIR_BUILD=$PWD
+DIR_SRC=$(cd $(dirname "$0")/../..; pwd -P)
+DIR_REPORT="$DIR_BUILD/sanitizers/report_$(date +"%y-%m-%d-%H%M%S")"
+DIR_GTEST_BUILD="$DIR_BUILD/gtest"
+DIR_GTEST_INSTALL="$DIR_GTEST_BUILD/googletest/install"
+GTEST_SETUP="$DIR_SRC/ci/travis/setup_gtest.sh"
+REPORT_BUILD="$DIR_REPORT/report_build"
+REPORT_CTEST="$DIR_REPORT/ctest_log"
+
+# Set a default sanitizer if envvar is not set
+if [ -z "$SANITIZER" ]
+then
+    SANITIZER="ASAN"
+    echo "INFO: environment variable SANITIZER is not set. Will be using by default $SANITIZER"
+fi
+
+# ASAN: Address Sanitizer
+ASAN="-fsanitize=address "
+ASAN_FLAGS="-fsanitize-address-use-after-scope -fsanitize-recover=address -fno-omit-frame-pointer -fno-common "
+# LSAN: Leak Sanitizer
+LSAN="-fsanitize=leak "
+LSAN_FLAGS=""
+# UBSAN: UndefinedBehaviour Sanitizer
+UBSAN="-fsanitize=undefined "
+UBSAN_FLAGS="-fno-omit-frame-pointer "
+# Compile time out of bounds checks. See more: https://fuzzing-project.org/tutorial-cflags.html
+EXTRA_FLAGS="-Wformat -Werror=format-security -Werror=array-bounds "
+
+FLAGS=${SANITIZER}"_FLAGS"
+SANITIZER_LDFLAGS="${!SANITIZER}"
+SANITIZER_FLAGS="${!SANITIZER}${!FLAGS}"
+SANITIZER_FLAGS+="$EXTRA_FLAGS"
+SANITIZER_FLAGS+="-shared-libasan "
+# NOTE: option '-shared-libasan' for link shared library with:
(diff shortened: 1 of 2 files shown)


tests/sanitizers/examples/asan_01.cc
```diff
@@ -0,0 +1,5 @@
+int main(int argc, char **argv) {
+  int *array = new int[100];
+  delete [] array;
+  return array[argc];  // BOOM
+}
```
bitmouse, line 5: ?????

tests/sanitizers/examples/asan_02.cc
```diff
@@ -0,0 +1,10 @@
+#include <stdlib.h>
+#include <stdio.h>
+#define N 10
+int main(void) {
+  char s[N]  = "123456789";
+  for (int i = 0; i <= N; i++)   // Loop bound goes past the end of the array
+    printf ("%c", s[i]);
+  printf("\n");
+  return 0;
+}
```
bitmouse, line 10: ???????

tests/sanitizers/examples/asan_03.cc
```diff
@@ -0,0 +1,9 @@
+#include <stdlib.h>
+#include <stdio.h>
+int main(void) {
+  char *s = (char*)malloc(sizeof(char) * 10);
+  free(s);
+  free(s);  // DOUBLE FREE
+  printf("%s", s);
+  return 0;
+}
```
bitmouse, line 9: Why you add these?

bitmouse (changes requested):

tests/sanitizers/examples/asan_01.cc
pfagner: This is test case for sanitizer

bitmouse: We don't test sanitizers. We use them.

tests/sanitizers/test_sanitizers.sh
```diff
@@ -0,0 +1,149 @@
...
+# │  │   └─ test_sanitizers.sh*   <--- executable script
+
+function outerdir(){
+   echo "$1" | sed -E 's@/[^.]+.*@/@'
+}
+DIR_BUILD=$PWD
+DIR_OUTER=$(outerdir "$0")
+if [ "$DIR_OUTER" = "./" ]
+then
+     echo "The script must be run from the 'facelift/build'"
```
jacky309, line 36: we do not want that. See my remark on the other PR.

pfagner: Okay.

bitmouse (changes requested):
I am unable to generate any useful output using this instruction. I can see only build output and this is not what I expect.
Please generate report from e.g. running tests. Suggestions review.

tests/sanitizers/test_sanitizers.sh
```diff
@@ -0,0 +1,98 @@
...
+SANITIZER_LDFLAGS="${!SANITIZER}"
+SANITIZER_FLAGS="${!SANITIZER}${!FLAGS}"
+SANITIZER_FLAGS+="$EXTRA_FLAGS"
+SANITIZER_FLAGS+="-shared-libasan "
+# NOTE: option '-shared-libasan' for link shared library with:
+# ASAN as a shared library. See more: https://github.com/google/sanitizers/wiki/AddressSanitizerAsDso
+# UBSAN as a shared library. See more: https://reviews.llvm.org/D38525
+
+CXX_VERSION_OUTPUT=$(clang --version)
+CXX_VERSION=$(echo $CXX_VERSION_OUTPUT | cut -d"-" -f1 | cut -d" " -f4)
```
bitmouse, line 43: This prints nothing
$ echo $CXX_VERSION



$ echo $CXX_VERSION_OUTPUT
clang version 9.0.0-2 (tags/RELEASE_900/final) Target: x86_64-pc-linux-gnu Thread model: posix InstalledDir: /usr/bin


try 
clang --version | head -n 1 | grep -o -E "[[:digit:]].[[:digit:]].[[:digit:]]" | uniq | sort

tests/sanitizers/test_sanitizers.sh
```diff
@@ -0,0 +1,98 @@
...
+    ;;
+    *)
+       echo "Sanitizer not specified"
+    ;;
+esac
+
+# Run tests
+case "${SANITIZER}" in
+  "ASAN")
+    PATH_LIBCLANG_RT=$(locate $LIBCLANG_RT_ASAN | grep clang | grep $CXX_VERSION)
```
bitmouse, line 80: because $CXX_VERSION is empty in my case, this fails in this line.

bitmouse (changes requested):

tests/sanitizers/test_sanitizers.sh
pfagner: The expression contain regex that have a mistake, but it is not a problem.
True way: is read source code (#define ...), but not an interpretation output for human.

bitmouse (approved):

PR #297, opened 2020-08-20
Add 'llvm-cov gcov' reads coverage data and displays the coverage information

pfagner:
…the coverage information

### DIFF
--- a/tests/coverage_lcov/read.me
+++ b/tests/coverage_lcov/read.me
@@ -1,0 +1,42 @@
+# The LLVM-COV -- is a tool shows code coverage information for programs that are instrumented to emit profile data.
+ The 'llvm-cov gcov' tool reads code coverage data files and displays
+ the coverage information for a specified source file.
+ It is compatible with the GCOV tool from version 4.2 of GCC and may also be compatible with some later versions of GCOV.
+
+# The GCOV -- is a source code coverage analysis and statement-by-statement profiling tool.
+ Gcov generates exact counts of the number of times each statement in a program is executed and annotates source code to add instrumentation.
+ Provides the following details:
+ - How often each line of code executes
+ - What lines of code are actually executed
+ - How much computing time each section of code uses
+
+# The LCOV -- is a graphical front-end for GCC's coverage testing tool GCOV.
+ It collects gcov data for multiple source files and creates HTML pages containing the source code annotated with coverage information.
+ It also adds overview pages for easy navigation within the file structure. LCOV supports statement, function and branch coverage measurement.
+
+# source code:
+ LCOV:     http://ltp.sourceforge.net/coverage/lcov.php
+ LLVM-COV: git clone http://llvm.org/git/llvm.git
+# docs:
+ LCOV:     http://ltp.sourceforge.net/coverage/lcov/lcov.1.php
+ LLVM-COV: https://llvm.org/docs/CommandGuide/llvm-cov.html
+# install using apt-get:
+ LLVM-COV: when installing using apt-get install llvm-<NUMBER>, llvm-cov will be installed with a suffix of the version number as well.
+ $ apt-get install llvm-<NUMBER>
+ LCOV:
+ $ apt-get install lcov
+
+# CAUTION: Create a wrapper script (patch) for LCOV
+ LLVM has a fast development speed, thus the llvm-cov command line options have been changed, and lcov can no longer recognize it.
+ Thus, we have to create a script llvm-gcov.sh to workaround the problem:
+ See more: http://logan.tw/posts/2015/04/28/check-code-coverage-with-clang-and-lcov/#create-a-wrapper-script-for-lcov
+           https://users.rust-lang.org/t/howto-generating-a-branch-coverage-report/8524
+ #!/bin/bash
+ exec llvm-cov gcov "$@"
+ $ chmod +x llvm-gcov.sh
+
+# How to use
+ Run script:
+   ~/facelift/build$ tests/coverage_lcov/test_lcov.sh
+ Results of coverage will be available in the folder
+   ~/facelift/build$ lcov/report_lcov_yy-mm-dd-hms
--- a/tests/coverage_lcov/test_lcov.sh
+++ b/tests/coverage_lcov/test_lcov.sh
@@ -1,0 +1,73 @@
+#!/bin/bash
+set -eu
+# for info see 'read.me'
+
+function outerdir(){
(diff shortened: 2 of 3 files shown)


tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,99 @@
...
+# ├─ tests/
+# │  ├─ asyncfunctions/
+# │  ├─ benchmarking/
+# │  ├─ cmake/
+# │  ├─ CMakeLists.txt
+# │  ├─ combined/
+# │  ├─ coverage_lcov/
+# │  │   ├─ llvm-gcov.sh    <--- patch for LCOV
+# │  │   ├─ read.me
+# │  │   └─ test_lcov.sh*   <--- executable script
```
bitmouse, line 28: _-- NOTE: Drawing will be removed after approval script_

Note: this review will not get an approval with this comment and this note.

tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,99 @@
...
+
+function outerdir(){
+   echo "$1" | sed -E 's@/[^.]+.*@/@'
+}
```
bitmouse, line 32: why do we need this?

tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,99 @@
...
+DIR_GTEST="$DIR_BUILD/gtest"
+DIR_GTEST_INSTALL="$DIR_GTEST/googletest/install"
+GTEST_SETUP="$DIR_CI/setup_gtest.sh"
+DATE_TIME=$(date +"%y-%m-%d-%H%M%S")
+DIR_REPORT="$DIR_BUILD/lcov/report_coverage_${DATE_TIME}"
+REPORT_COVERAGE="$DIR_REPORT/index.html"
+REPORT_CTEST="$DIR_REPORT/ctest.log"
+COVERAGE_FILE="${DIR_BUILD}/coverage.info"
+COVERAGE_OUT_FILE="${DIR_BUILD}/coverage_out.info"
+NUMBER_THREADS="-j 4"
```
bitmouse, line 51: NUMBER_THREADS=4

Please add "-j" somewhere else, or rename variable

tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,99 @@
...
+  then
+    mkdir -p $DIR_GTEST
+    pushd $DIR_GTEST
+    # Setup test
+    . $GTEST_SETUP
+    popd
+fi
+CMAKE_SYSTEM_PREFIX_PATH="$DIR_GTEST_INSTALL"
+# Build facelift with flags
+cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=${COMPILER} -DCMAKE_EXE_LINKER_FLAGS="-lgcov --coverage" -DCMAKE_CXX_FLAGS="-fprofile-arcs -ftest-coverage" -DFACELIFT_BUILD_TESTS=ON -DCMAKE_SYSTEM_PREFIX_PATH=$CMAKE_SYSTEM_PREFIX_PATH $DIR_SRC && cmake --build ${DIR_BUILD} -- $NUMBER_THREADS
```
bitmouse, line 65: what is the ${COMPILER}?
Do you create coverage report for each compiler? We don't want to do that. Pick one.

bitmouse (changes requested):

tests/coverage_lcov/test_lcov.sh
pfagner: Okay. I will remove this drawing.

tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,73 @@
...
+# for info see 'read.me'
+
+function outerdir(){
+   echo "$1" | sed -E 's@/[^.]+.*@/@'
+}
+DIR_BUILD=$PWD
+DIR_OUTER=$(outerdir "$0")
+if [ "$DIR_OUTER" = "./" ]
+then
+     echo "The script must be run from the 'facelift/build'"
```
jacky309, line 12: The script should be runnable from anywhere, especially from folders which are outside of the source folder since we do not want to write build files into the source folder.

tests/coverage_lcov/test_lcov.sh
```diff
@@ -0,0 +1,73 @@
...
+     exit 1
+fi
+DIR_SRC=$(cd "$DIR_OUTER"; pwd -P)
```
jacky309, line 15: DIR_SRC=$(cd $(dirname "$0")/..; pwd -P)

jacky309: then you can remove that outerdir stuff....

jacky309 (changes requested):

tests/coverage_lcov/test_lcov.sh
pfagner: We have already discussed the "related path .." and how to deduce the source path from the dirname.
If the this variant you proposed suits you, then this is not a problem.
Thank you.

jacky309: It is simple and it does work, so why would we need anything else ? Why did you change that in the first place ?

tests/coverage_lcov/read.me
```diff
@@ -0,0 +1,42 @@
+# The LLVM-COV -- is a tool shows code coverage information for programs that are instrumented to emit profile data.
```
jacky309, line 1: Please do not insert lcov doc here. Simply provide a link to it.

tests/coverage_lcov/test_lcov.sh
pfagner: Q: Why did you change that in the first place?
A: I did to conclusion from your short sentences that is required is more "smart" solution for non trivial decision with traverse of folders.
J'ai pris vos paroles trop au sérieux. Vous vouliez le plus simple, je voulais le meilleur. Ne vous fâchez pas, tout ira bien de toute façon.

tests/coverage_lcov/read.me
pfagner: Okay.

pfagner:
This commit is closed due to no need to use two LLVM-COV + GCOV tools. Instead of it will be use LLVM-COV only:
https://github.com/Pelagicore/facelift/pull/302
All notes will be implement in related tool/commit.

PR #299, opened 2020-08-28
Missing virtual dtors

pfagner:


bitmouse (changes requested):
please move '=default' to .cpp if possible. This will allow to avoid future problems.

bitmouse (approved):

PR #303, opened 2020-09-02
Standardize facelift dbus communication

idleroamer:
[WIP] Standardize facelift dbus communication
Change-Id: Ifd09ce0e4b16a6a060d2a2634adc614f56b3a139


jacky309 (changes requested):
We went away from using the "native" DBus serialization a while ago, for 2 reasons:
    - Performance
    - We have reached the maximum number of arguments which can be written in a message.

AFAIU, your commit reverts that change, which is a bad idea.

bitmouse:
> We went away from using the "native" DBus serialization a while ago, for 2 reasons:
> - Performance
> - We have reached the maximum number of arguments which can be written in a message.
> 
> AFAIU, your commit reverts that change, which is a bad idea.

1. This commit is PoC for further discussion.
2. If what you say is true, then we should probably drop DBus and use other standard. Either we should use common standards or different tool/protocol.
3. I don't really believe in performance argument. Benchmarks to be done (I was not aware of this argument up to now). 
4. If we have to drop native DBus then we should probably use some existing tool that allows us to communicate with other languages/systems, e.g. flatbuffers. But then we should probably use sockets directly.

idleroamer:
I could only find the following limitations:
* The maximum length of a signature is 255. 
* The maximum depth of container type nesting is 32 array type codes and 32 open parentheses. This implies that the maximum total depth of recursion is 64, for an "array of array of array of ... struct of struct of struct of ..." where there are 32 array and 32 struct. 
Are you referring to those limitations or others? Because IMHO those are generous limits!

bitmouse:
> - We have reached the maximum number of arguments which can be written in a message.

as noticed this was most probably due to incorrect serialization of lists of structures, so if done properly it should not be the case

jacky309:
> > We went away from using the "native" DBus serialization a while ago, for 2 reasons:
> > 
> > * Performance
> > * We have reached the maximum number of arguments which can be written in a message.
> > 
> > AFAIU, your commit reverts that change, which is a bad idea.
> 
> 1. This commit is PoC for further discussion and testing.
> 2. If what you say is true, then we should probably drop DBus and use other standard. Either we should use common standards or different tool/protocol.
Only the serialization part of DBus is not used in a classical manner. We need an IPC to deliver our messages, which is where DBus useful for us.

> 3. I don't really believe in performance argument. Benchmarks to be done (I was not aware of this argument up to now).
Feel free to write a benchmark if you really need that in order to be convinced.

> 4. If we have to drop native DBus then we should probably use some existing tool that allows us to communicate with other languages/systems, e.g. flatbuffers. But then we should probably use sockets directly.
flatbuffers is a serialization library, it provides no IPC.

jacky309:
> > * We have reached the maximum number of arguments which can be written in a message.
> 
> as noticed this was most probably due to incorrect serialization of lists of structures, so if done properly it should not be the case

jacky309:
> > > * We have reached the maximum number of arguments which can be written in a message.
> > 
> > 
> > as noticed this was most probably due to incorrect serialization of lists of structures, so if done properly it should not be the case

The problem is that Qt is missing an API which would be needed to serialize structures "properly". I have spent a lot of time trying to find a workaround. Maybe that API is available in a more recent version of Qt though.

idleroamer:
IDK about the time of facelift initial implementation but as Qt5.12 the API covers all requirements.
My initial benchmarking doesn't show any regression (as I suspected the overhead in dbus is communication not the property marshal/unmarshalling).

jacky309:
What kind of data did you use in your benchmark ? The overhead introduced by the serialisation is not significant if only a small amount of data is being transferred, so please adjust your benchmark to include large data such as long lists. Also, the registration of all those types in QDBus type system probably has a signification performance cost at startup.
Your change introduces a dependency to DBus into the "local" and "common" IPC components, which is not acceptable.
I did not dig into the details of your changes but I would not be surprised if the support for sub-interfaces over IPC was lost (especially sub-interfaces in structures).
Also, Facelift enables the change of multiple properties at once in an atomic way, which can be important if the values of those properties are related and should be consistent when being exposed. When exposed over IPC, the consistency should be kept. I suspect your change to break that consistency.

If you resolve the merge conflict, we can see if some tests fail, but keep in mind that our test coverage is still incomplete.

idleroamer:
We are doing further benchmarking and will include the QDbus type registerations as well.
I tried not to break the "interfaces", even though the whole encapsulation is disaster, especially there are many inconsistency between list, map of "interfaces" and stand-alone ones. We need to follow up on that in some other effort.
The simultaneous change of multiple properties doesn't make sense to me at all. If a set of properties are dependent and represent state of any object, following Object-Oriented design, they should be packed into a struct, beside nobody expects this as well and it is not dbus conformant.

bitmouse:
> Also, the registration of all those types in QDBus type system probably has a signification performance cost at startup.

Could you explain better what you mean? 

> I did not dig into the details of your changes but I would not be surprised if the support for sub-interfaces over IPC was lost (especially sub-interfaces in structures).

To be verified. Please ask for checking instead of personal comments in the future.


> Facelift enables the change of multiple properties at once in an atomic way, 

yeah... atomic... right... I have never seen any kind of mutex, etc. in facelift. Not even saying about std::atomic.

jacky309:
> > Also, the registration of all those types in QDBus type system probably has a signification performance cost at startup.
> 
> Could you explain better what you mean?

Just look at the amount of additional code for type registrations. The registration template methods typically expands into large portions of binary code.

> 
> > I did not dig into the details of your changes but I would not be surprised if the support for sub-interfaces over IPC was lost (especially sub-interfaces in structures).
> 
> To be verified. Please ask for checking instead of personal comments in the future.

There is nothing personal in my comment ! I noticed large portions of code being removed/replaced, including the code which is responsible for the registration of sub-interfaces. This stuff can be broken without noticing at first. This needs to be checked, ideally with a new test case, if this feature is not covered yet.

> 
> > Facelift enables the change of multiple properties at once in an atomic way,
> 
> yeah... atomic... right... I have never seen any kind of mutex, etc. in facelift. Not even saying about std::atomic.

jacky309:
> We are doing further benchmarking and will include the QDbus type registerations as well.
> I tried not to break the "interfaces", even though the whole encapsulation is disaster, especially there are many inconsistency between list, map of "interfaces" and stand-alone ones. We need to follow up on that in some other effort.
> The simultaneous change of multiple properties doesn't make sense to me at all. If a set of properties are dependent and represent state of any object, following Object-Oriented design, they should be packed into a struct, beside nobody expects this as well and it is not dbus conformant.

To make sure we are on the same page, let's consider that example:

interface TunerService {
    int currentFrequency;
    int maximumFrequency;
    int minimumFrequency;
}

Now, please consider the "AM to FM band switching" use case, where those 3 properties are changed. If you first change the currentFrequency and issue the "currentFrequency changed" signal, you possibly expose an inconsistent state to the slots, since the value of the currentFrequency property is possibly outside of the validity range.
A correct way to address this issue is to change all property values first, and then start triggering the corresponding change signals. By doing so, no inconsistent state is exposed to any slot. Facelift's IPC layer preserves that consistency, by making a distinction between those 2 events: "property X changed" and "XChanged signal is triggered".

I agree that moving all those "related properties" into a single structure property would be a way to ensure that the consistency is preserved whatever happens, but this is a workaround which has the following problems:
    - It results in a weird interface definition since the fields are not grouped in a natural or logical way, but rather in such a way that the consistency is insured.
    - When a single value has changed, multiple additional fields are also sent over IPC, if they belong to the same structure, even if only one field has changed. This can be an issue with large properties (bitmap or long lists).

bitmouse:
> 
> interface TunerInterface {
> int currentFrequency;
> int maximumFrequency;
> int minimumFrequency;
> }
> 
> Now, please consider the "AM to FM band switching" use case, where those 3 properties are changed. If you first change the currentFrequency and issue the "currentFrequency changed" signal, you possibly expose an inconsistent state to the slots, since the value of the currentFrequency property is possibly outside of the validity range, since the latter has not been changed yet.
> A correct way to address this issue is to change all property values first, and then start triggering the corresponding change signals. By doing so, no inconsistent state is exposed to any slot. Facelift's IPC layer preserves that consistency, by making a distinction between those 2 events: "property X changed" and "XChanged signal is triggered".
> 

I think this is wrong already on the concept level. At first, the interface itself should be stateless, so it should have no properties. At second, these should be grouped into a structure. The biggest benefit of grouping them into the structure is that we start following the object oriented programming paradigm. Even C language provided structures for the inseparable data, so why do we?

> I agree that moving all those "related properties" into a single structure property would be a way to ensure that the consistency is preserved whatever happens, but this is a workaround which has the following problems:
> - It results in a weird interface definition since the fields are not grouped in a natural or logical way, but rather in such a way that the consistency is insured.

If we have the proper (stateless) interface, it is not a workaround. If the fields are not grouped in the natural or logical way (structured), then this is the project design problem, not the RPC framework that project uses.

> - When a single value has changed, multiple additional fields are also sent over IPC, if they belong to the same structure, even if only one field has changed. This can be an issue with large properties (bitmap or long lists).

If this is a problem, I would suggest to refine the design of the "interface" or structures from project side, because something is wrong at this level. 


If something is unclear about grouping: imagine that you have to send Point2D. What is the proper way:

interface Point {
int x;
int y;
}

or 

interface Point {
Point2D getPoint();
void setPoint(Point2D p);
}

struct Point2D {
int x;
int y;
}

Even if only single field has changed, this is now completely different point.

idleroamer:
@jacky309 We were following up on dbus limit and for some huge structure the 255 character limit of the signature could be an issue. Although we can switching to old binary data transfer for those structs when annotated with "@serializable:true", I am not fully sure what is the main purpose of that annonation but since previously all structs were send like it is no backward compatibility issue.
Please have a look at:
https://github.com/Pelagicore/facelift/pull/303/commits/c1ad3438471f9250ccf8afe5cfe901efd34d540b

jacky309:
> @jacky309 We were following up on dbus limit and for some huge structure the 255 character limit of the signature could be an issue. Although we can switching to old binary data transfer for those structs when annotated with "@serializable:true", I am not fully sure what is the main purpose of that annonation but since previously all structs were send like it is no backward compatibility issue.
> Please have a look at:
> [c1ad343](https://github.com/Pelagicore/facelift/commit/c1ad3438471f9250ccf8afe5cfe901efd34d540b)

That annotation is documented in the annotations page. It adds the serialize/deserialize methods to the structure API, which is completely unrelated to the IPC layer. Reusing that annotation to specify how the structure is serialized over DBus-IPC sounds like a misuse.
I would rather introduce a new annotation to enable an interface to be exposed in a "native DBus" way and I would keep the current serialization approach as default. I still do not know exactly what concrete problem you are trying to solve with that pull request, TBH. Maybe you can clarify that ? If the goal is to improve the interoperability issues, I assume that only some specific interfaces are concerned.

idleroamer:
Well as you put it the issue is interoperability, we would like to use facelift in combination with other dbus frameworks in other languages.
With this PR it is possible to use other frameworks as well (godbus), idk what do you mean by some specific interfaces.
We can introduce a new annotation should not be hard. But I see the default behaviour otherway around, all communication should be in "native dbus" (since facelift used dbus and intuitively everybody expects that) unless the user defined it otherwise (due to huge structure signature or otherwise).

jacky309:
> > interface TunerInterface {
> > int currentFrequency;
> > int maximumFrequency;
> > int minimumFrequency;
> > }
> > Now, please consider the "AM to FM band switching" use case, where those 3 properties are changed. If you first change the currentFrequency and issue the "currentFrequency changed" signal, you possibly expose an inconsistent state to the slots, since the value of the currentFrequency property is possibly outside of the validity range, since the latter has not been changed yet.
> > A correct way to address this issue is to change all property values first, and then start triggering the corresponding change signals. By doing so, no inconsistent state is exposed to any slot. Facelift's IPC layer preserves that consistency, by making a distinction between those 2 events: "property X changed" and "XChanged signal is triggered".
> 
> I think this is wrong already on the concept level. .....

Why not sticking to the example which I suggested ? :
Again, I have a radio tuner device, which is a stateful device. Its state is made of 2 things:
   - a selected frequency
   - a selected wave band
Now, if I want to use Facelift to expose an interface which enables someone to interact with my device, I would likely come up with the following QFace definition, which provides some additional properties for maximum and minimum frequencies:

enum WaveBand {
  AM,
  FM
}

interface TunerInterface {
   int currentFrequency;
   readonly int maximumFrequency;
   readonly int minimumFrequency;
   WaveBand waveband; 
 }

What is fundamentally wrong with that interface ? Why ? What is not object-oriented here, if all the properties belong to an object (the interface) ?

How would your QFace interface look like, for the same tuner, with the same feature set ?
It is important that we have a common understanding of what a good interface looks like, to understand what Facelift should provide.

jacky309:
> Well as you put it the issue is interoperability, we would like to use facelift in combination with other dbus frameworks in other languages.
> With this PR it is possible to use other frameworks as well (godbus). IDK what do you mean by some specific interfaces is only needed to enable interoperability!

I assume that most QFace interfaces defined in a product using facelift do not require the interoperability feature. Typically, only a couple of them would really require the interoperability annotation.

> We can introduce a new annotation should not be hard. But I see the default behaviour otherway around, all communication should be in "native dbus" (since facelift used dbus and intuitively everybody expects that) unless the user defined it otherwise (due to huge structure signature or otherwise).

The current implementation does not have the limitations which your change introduces and I assume it is faster and generates a more compact code (to be confirmed). The current implementation can also easily be improved to get multiple consecutive property change signals to be squashed into a single dbus message. AFAIK, your approach does not allow that.

I understand that the interoperability feature is useful and I can appreciate the work you've done, but the price to pay should definitely be considered.

bitmouse:
> 
> Why not sticking to the example which I suggested ? :
> Again, I have a radio tuner device, which is a stateful device. Its state is made of 2 things:
> 
>     * a selected frequency
> 
>     * a selected wave band
>       Now, if I want to use Facelift to expose an interface which enables someone to interact with my device, I would likely come up with the following QFace definition, which provides some additional properties for maximum and minimum frequencies:
> 
> 
> enum WaveBand {
> AM,
> FM
> }
> 
> interface TunerInterface {
> int currentFrequency;
> readonly int maximumFrequency;
> readonly int minimumFrequency;
> WaveBand waveband;
> }
> 
> What is fundamentally wrong with that interface ? Why ? What is not object-oriented here, if all the properties belong to an object (the interface) ?

As mentioned before it created state of the interface. So this is not right, the proper interface is state agnostic (stateless). That is the first fail of this kind of "interface" - it has properties. You have created remote object instead of interface. 
However this still would be acceptable if they were organized properly. 

> 
> How would your QFace interface look like, for the same tuner, with the same feature set ?
> It is important that we have a common understanding of what a good interface looks like, to understand what Facelift should provide.

Please consider this:

interface TunerInterface {
  int currentFrequency;
  BandInfo bandInfo;
}

struct BandInfo {
  readonly int  maximumFrequency;
  readonly int minimumFrequency;
  WaveBand waveband;
}

The currentFrequency can change independently to other properties, but the other three are tightly coupled. If we change bandInfo now, we must probably change currentFrequency, but this is developer responsibility now to do it right, not the, so called, interface (remote object in fact).

But with further improvement I would suggest (BandInfo structure as above).

interface Tuner { // you do not have to tell it is interface in the name. Using Hungarian notation and its modifications is really bad idea
  int getCurrentFrequency();
  void setCurrentFrequency(int freq); // or bool, where bool is a success status
  BandInfo getBandInfo(); 
  void setWaveBand(WaveBand wb);
  signal currentFrequencyChanged();
  signal bandChanged();
}

This makes the interface design right, interface is stateless. Using polymorphism in this case makes sense. Also it could be generated to other programming languages as an interface. 


> The current implementation does not have the limitations which your change introduces and I assume it is faster and generates a more compact code (to be confirmed). The current implementation can also easily be improved to get multiple consecutive property change signals to be squashed into a single dbus message. AFAIK, your approach does not allow that.

This change actually removes limitation to facelift framework in DBus usage. Anyway there are other problems like service registry which is in fact useless and most probably can be removed with minimal effort.

idleroamer:
I will try to summarize all points here:

Purpose of PR:

Facelift is not a standard dbus framework and hence not interoperable due to following reasons.
* Signatures of properties/methods/signals are missing or are completely wrong
* “org.freedesktop.Dbus.Properties interface” is not utilized
* All signals are send via generic “SignalTriggered” 
* Methods doesn’t have right signature
All the above makes Facelift impossible to combine in form of server/client schema with other dbus tools/frameworks.

Approach: Refactor the ipc layer in following manner
* Utilize QtDBus tools to generate right signature for structs
* Utilize “org.freedesktop.Dbus.Properties interface” as the standard interface for property exchange
* For each signal use the corresponding signal signature not the generic “SignalTriggered”

Concerns about the PR

* DBus limitation:
    • The struct in struct limitation of 32 by 32.
    • The maximum length of a signature is 255. 
Remedy: Annotation: @serializeOverIPC turns the struct into byte array form over the dbus interfaces. Similar to previous treatment of structs.

Default Behavior: Intuitively and obviously this is not the default behavior of struct, the struct needs to follow dbus standard.
If developer runs to such limitation in their code-base, they really need to rethink their design. 255 is really long signature (in whole carbon-ui two auto-generated if1qface structs run into this limit which IMHO are badly design as well).


* Atomic PropertiesChanged of multiple properties:
Dbus standard allow sending a “PropertiesChanged” with multiple values as well. 
Event though AFAIK facelift doesn’t support this feature as current status but anyhow PR does **not** block future features in this direction.

org.freedesktop.DBus.Properties.PropertiesChanged (STRING interface_name, DICT<STRING,VARIANT> changed_properties, ARRAY<STRING> invalidated_properties);



* Performance regression
Preliminary results show no regression in case of 10000 method calls of methods with different range of  method signature (input and output struct) from couple of fileds to ~200 fields in struct. 

Please reachout for .pref files.

jacky309:
> > Why not sticking to the example which I suggested ? :
> > Again, I have a radio tuner device, which is a stateful device. Its state is made of 2 things:
> > ```
> > * a selected frequency
> > 
> > * a selected wave band
> >   Now, if I want to use Facelift to expose an interface which enables someone to interact with my device, I would likely come up with the following QFace definition, which provides some additional properties for maximum and minimum frequencies:
> > ```
> > 
> > 
> > enum WaveBand {
> > AM,
> > FM
> > }
> > interface TunerInterface {
> > int currentFrequency;
> > readonly int maximumFrequency;
> > readonly int minimumFrequency;
> > WaveBand waveband;
> > }
> > What is fundamentally wrong with that interface ? Why ? What is not object-oriented here, if all the properties belong to an object (the interface) ?
> 
> As mentioned before it created state of the interface. So this is not right, the proper interface is state agnostic (stateless). That is the first fail of this kind of "interface" - it has properties. You have created remote object instead of interface.
> However this still would be acceptable if they were organized properly.
> 
> > How would your QFace interface look like, for the same tuner, with the same feature set ?
> > It is important that we have a common understanding of what a good interface looks like, to understand what Facelift should provide.
> 
> Please consider this:
> 
> interface TunerInterface {
> int currentFrequency;
> BandInfo bandInfo;
> }
> 
> struct BandInfo {
> readonly int maximumFrequency;
> readonly int minimumFrequency;
> WaveBand waveband;
> }
> 
> The currentFrequency can change independently to other properties, but the other three are tightly coupled. If we change bandInfo now, we must probably change currentFrequency, but this is developer responsibility now to do it right, not the, so called, interface (remote object in fact).
> 
> But with further improvement I would suggest (BandInfo structure as above).
> 
> interface Tuner { // you do not have to tell it is interface in the name. Using Hungarian notation and its modifications is really bad idea
> int getCurrentFrequency();
> void setCurrentFrequency(int freq); // or bool, where bool is a success status
> BandInfo getBandInfo();
> void setWaveBand(WaveBand wb);
> signal currentFrequencyChanged();
> signal bandChanged();
> }
> 
> This makes the interface design right, interface is stateless. Using polymorphism in this case makes sense. Also it could be generated to other programming languages as an interface.
> 
> > The current implementation does not have the limitations which your change introduces and I assume it is faster and generates a more compact code (to be confirmed). The current implementation can also easily be improved to get multiple consecutive property change signals to be squashed into a single dbus message. AFAIK, your approach does not allow that.
> 
> This change actually removes limitation to facelift framework in DBus usage. Anyway there are other problems like service registry which is in fact useless and most probably can be removed with minimal effort.

So you are suggesting that users of Facelift stop defining interfaces which contain properties. That would mean that :
    - Interfaces are not QML-friendly anymore. QML shines at handling properties (especially property bindings). One of the primary goals of Facelift is to help exposing interfaces to QML.
    - A major performance penalty is introduced. Property values now have to be requested one by one, by calling the property getters.
    - Worst of all: by the time the property getters have returned, the values which have been received by the client are not guaranteed to be consistent, since it could be that the server side changed some properties in between. There is no reliable way, at any point time, for the client to know if the received values represent a state which ever existed on the server side.

Your interface is a perfect example of what NOT to do, especially as far as defining interfaces to be used over IPC.
Our RPC framework has to ensure that the state replicated on the client side is a state which existed on the server side and to minimize the replication time. We do not want any user code to be required to realize that reliable replication.
AFAIK, the only way to replicate the state reliably is to be able to transport multiple property updates together in a single IPC message, which is what facelift currently does, and we should not break that.

By turning formal properties into methods/signals, you are just hiding the real nature of the interface, which is stateful, and you simply introduce new problems.

idleroamer:
As I pointed out in the last commit, this PR doesn't put any limit on how many properties can be changed atomically.
org.freedesktop.DBus.Properties.PropertiesChanged (STRING interface_name, **DICT<STRING,VARIANT> changed_properties**, ARRAY invalidated_properties);

So please let's discuss scope of the PR, we needs these changes ASAP.

jacky309:
Added new test : https://github.com/Pelagicore/facelift/pull/305
Could you please check that your change does not break that test ?
Could you please also squash your commits ?

codegen/facelift/templates/Struct.template.cpp
```diff
@@ -126,5 +131,77 @@ QString {{struct.name}}::toString() const
     return toStringWithFields(CLASS_ID, FIELD_NAMES);
 }
 
+#ifdef DBUS_IPC_ENABLED
```
jacky309: The structure definition headers are and should remain free from any IPC-related code.
Please create new headers for the DBus specific serialization code.

src/ipc/dbus/DBusIPCCommon.h
```diff
@@ -36,15 +36,17 @@
...
 
+
 namespace facelift {
 namespace dbus {
 using namespace facelift;
 
+
 struct FaceliftIPCLibDBus_EXPORT DBusIPCCommon {
-    static constexpr const char *GET_PROPERTIES_MESSAGE_NAME = "GetAllProperties";
+    static constexpr const char *GET_ALL_PROPERTIES = "GetAll";
```
jacky309: Please keep the suffixes such as _MESSAGE_NAME

src/ipc/ipc-common/IPCProxyModelProperty.h
```diff
@@ -30,6 +30,9 @@
 #pragma once
 
 #include <QObject>
+#include <QtDBus>
```
jacky309: no DBus stuff in that header please.

jacky309:
"serializeOverIPC" is not a good name. Structures are always serialized to be sent over IPC...
I would rather have an annotation on an interface, rather than on the structure definition.

codegen/facelift/templates/Struct.template.cpp
roam3r: I don't get the rationale for moving it out. Facelift is a DBus framework so its structure should now how to register themselves as dbus types and beside there are #ifdef guards for IPC_DISABLED builds.

src/ipc/ipc-common/IPCProxyModelProperty.h
roam3r: The same here, what is the rationale for it? The facelift is based on DBus.

codegen/facelift/templates/Struct.template.cpp
jacky309: DBus is an implementation detail of Facelift's IPC layer, and Facelift can be built without DBus. No user code may depend on DBus being used or not, which means no definition related to DBus should be included when including a structure header. This is also a way to minimize the build time.

roam3r: There are ifdef guards for IPC_DISABLED so no QtDBus is included in that build variant.
If user build facelift with IPC_ENABLED obviously there will be IPC (QtDBus) related codes included, having those in header file or not has no effect on build time whatsoever. 

The following piece of code belongs logically to "struct" since it deals with struct's internal data structure and how to serialize it to QDBusArgument. So it can't and shouldn't moved to someplace else.

QDBusArgument &operator<<(QDBusArgument &argument, const {{struct.name}} &{{struct.name|lower}})

idleroamer:
> Added new test : #305
> Could you please check that your change does not break that test ?
> Could you please also squash your commits ?

The test passes.
I will squash commits for the last step.

idleroamer:
> "serializeOverIPC" is not a good name. Structures are always serialized to be sent over IPC...
> I would rather have an annotation on an interface, rather than on the structure definition.

switch to "treatAsByteArray".
The annotation is about each single struct not the interface itself. So some structs in interface can be send as byte array while others are still in native signature.

codegen/facelift/templates/IPCAdapter.template.cpp
```diff
@@ -39,10 +39,27 @@
...
+
+{% for enum in module.enums %}
+#include "{{enum.fullyQualifiedPath}}.h"
+{% endfor %}
+
 {% for property in interface.referencedInterfaceTypes %}
 #include "{{property.fullyQualifiedPath}}{% if generateAsyncProxy %}Async{% endif %}IPCDBusAdapter.h"
 {% endfor %}
+
+#define DBUS_MAXIMUM_SIGNATURE_LENGTH 255
```
jacky309: move into common header

codegen/facelift/templates/IPCAdapter.template.cpp
```diff
@@ -80,6 +97,25 @@ struct {{interfaceName}}IPCAdapter::Impl {
 {{interfaceName}}IPCAdapter::{{interfaceName}}IPCAdapter(QObject* parent) :
     BaseType(facelift::InterfaceManager::instance(), parent)
 {
+#ifdef DBUS_IPC_ENABLED
```
jacky309: all that DBus specific code belongs to {{interfaceName}}IPCDBusAdapter

roam3r: I generally agree, but the next step is to remove the whole "local" specialization of the IPC adapter/proxies.
Since QtDBus has a built-in mechanism to treat the local async messages differently (similar to what "local" does)
so I have tried to treat as homogeneous as possible for the next patch, meaning use the QtDBus way of communication which means we need the qDBusRegisterMetaType calls.

jacky309: The "local" communication is needed on the platforms where QtDBus is not available.

src/ipc/ipc-common/IPCProxyModelProperty.h
jacky309: Your understanding that Facelift is based on DBus is wrong. It happens to use DBus for its IPC layer currently, but it could very well use another kind of channel. There has been some work on using another IPC in the past.
We do not want any DBus stuff inside the "ipc-common" folder.

codegen/facelift/templates/Struct.template.cpp
jacky309: Check the existing code to see how a struct is serialized without having to define the serialization function as "friend".

codegen/facelift/templates/IPCAdapter.template.cpp
roam3r: Then on platforms with QtDBus support the "local" files should not be generated and facelift should relying on QtDBus.
But that is subject of another discussion.

codegen/facelift/templates/Struct.template.cpp
roam3r: I have already seen that IMHO wrong pattern which not only break the encapsulation of class by exposing all internal fields (you may as well define the facelift struct internal's public) via "asTuple" rather has a huge negative impact on performance. 
In my benchmarks construction of tuples where always one of the bottle-necks in IPC so I changed from std::tuple to https://github.com/taocpp/tuple/ and observed up to ~15-20% performance gain, probably higher gain still possible if one doesn't use tuples at all.
Anyhow I have switched back to the old "asTuple" way since I don't want to mix scopes of changes.

idleroamer:
@jacky309 Please have another look, all QtDBus stuff moved to DBus layer.

src/model/FaceliftEnum.h
```diff
@@ -39,6 +39,78 @@
 #include "FaceliftCommon.h"
 #include <QObject>
 #include <QTextStream>
+#include <QDataStream>
+#ifdef DBUS_IPC_ENABLED
+#include <QDBusArgument>
+
+template<typename T, typename TEnum = void>
+class QDBusEnumMarshal;
```
jacky309: no DBus here

src/ipc/local/LocalIPCServiceAdapterBase.h
```diff
@@ -111,8 +102,53 @@ class FaceliftIPCLocalLib_EXPORT LocalIPCServiceAdapterBase : public IPCServiceA
...
+    template<typename T, typename std::enable_if_t<!std::is_convertible<T, facelift::InterfaceBase*>::value && std::is_enum<T>::value, int> = 0>
+    QVariant castToVariant(const T& value) {
+        return QVariant::fromValue(static_cast<int>(value));
+    }
+
+    template<typename T, typename std::enable_if_t<std::is_convertible<T, facelift::InterfaceBase*>::value, int> = 0>
+    QVariant castToVariant(const T& value) {
+        DBusObjectPath  dbusObjectPath;
+        if (value != nullptr) {
+            dbusObjectPath = DBusObjectPath (getOrCreateAdapter<typename std::remove_pointer<T>::type::IPCLocalAdapterType>(value)->objectPath());
```
jacky309: no DBus

src/ipc/local/LocalIPCProxy.h
```diff
@@ -138,7 +104,47 @@ class LocalIPCProxy : public IPCProxyBase<InterfaceType>, protected LocalIPCProx
...
+
+    template<typename T, typename std::enable_if_t<std::is_convertible<T, facelift::InterfaceBase*>::value, int> = 0>
+    T castFromVariantSpecialized(HelperType<T>, const QVariant& value) {
+        return getOrCreateSubProxy<typename std::remove_pointer<T>::type::IPCLocalProxyType>(qvariant_cast<DBusObjectPath>(value));
+    }
+
+    template<typename T>
+    QMap<QString, T*> castFromVariantSpecialized(HelperType<QMap<QString, T*>>, const QVariant& value) {
+        QMap<QString, T*> ret;
+        auto objectPaths = qvariant_cast<QMap<QString, DBusObjectPath>>(value);
```
jacky309: no dbus

src/ipc/local/LocalIPCProxy.h
```diff
@@ -42,6 +42,8 @@
 
 namespace facelift {
 
+typedef QString DBusObjectPath;
```
jacky309: dbus

src/ipc/local/LocalIPCMessage.cpp
```diff
@@ -96,6 +86,17 @@ void LocalIPCMessage::copyRequestMessage(const LocalIPCMessage &other)
     }
 }
 
+QList<QVariant> LocalIPCMessage::arguments() const
+{
+    return m_data.m_arguments;
+}
+
+LocalIPCMessage &LocalIPCMessage::operator<<(const QVariant &arg)
```
jacky309, line 94: no conversion to QVariant is expected for local IPC

roam3r: It is the Qt standard way of treating different types generically. As opposed to strange pattern of facelift::structure and collecting all fields in a std::tuples and its overhead one should have used QVariant when working with Qt.
The next step would be to remove the std::tuple.

jacky309: QVariant provides a way to handle different types generically at runtime. We do not need that here, and we do not want to pay the price of the overhead which the usage of QVariant introduces. All types are known at build time, and we want to benefit from C++'s type safety to minimize the potential runtime failures. I am not aware of any overhead of std::tuple.

roam3r: std::tuple overhead:
The bottle-neck is the construction of std::tuples at facelift::structure, as I already wrote the benchmark shows replacing std::tuple with some more efficient types like [tao::tuple](https://github.com/taocpp/tuple/pulls) proves ~15% speed gain. So how much we would gain by not having any tuple at all?!
Feel free to do more benchmarks.
Rationale:
Obviously QVariant creates some overhead at this phase but it is absolutely necessary to treat the IPCMessages generically, given that QtDBus is laid out on QVariants.
Type safety:
Current implementation of serializing values to input/Output payload (QByteArray) when sending local/DBus messages doesn't provide any type safety that QVariant would take away. In contrast using QtDBus we have clearer methods/struct signature (as oppose to ByteArray) for cross-referencing.

jacky309: > std::tuple overhead:
> The bottle-neck is the construction of std::tuples at facelift::structure, as I already wrote the benchmark shows replacing std::tuple with some more efficient types like [tao::tuple](https://github.com/taocpp/tuple/pulls) proves ~15% speed gain. So how much we would gain by not having any tuple at all?!

Where can I find that benchmark ? AFAIS, the main goal of tao::tuple is to optimize compilation time.

> Feel free to do more benchmarks.
> Rationale:
> Obviously QVariant creates some overhead at this phase but it is absolutely necessary to treat the IPCMessages generically, given that QtDBus is laid out on QVariants.

Use QVariant where it is needed, which means in the DBus IPC classes, not anywhere else.

> Type safety:
> Current implementation of serializing values to input/Output payload (QByteArray) when sending local/DBus messages doesn't provide any type safety that QVariant would take away. In contrast using QtDBus we have clearer methods/struct signature (as oppose to ByteArray) for cross-referencing.

codegen/facelift/templates/Struct.template.cpp
```diff
@@ -95,6 +96,12 @@ void {{struct.name}}::deserialize(const QByteArray &array)
 
 {% endif %}
 
+{% if struct.treatAsByteArray %}
+bool {{struct.name}}::treatAsByteArray()
```
jacky309: The fact that a structure is serialized as a byte array or not is not a property of the structure itself. It belongs to the IPC layer so it has nothing to do in the definition of the structure itself.

src/ipc/local/LocalIPCServiceAdapter.h
```diff
@@ -91,6 +91,12 @@ class LocalIPCServiceAdapter : public LocalIPCServiceAdapterBase
         registerService(objectPath, static_cast<ServiceType *>(serverObject));  // TODO: get rid of that cast
     }
 
+    template<typename T>
+    const char* typeToSignature() const
```
jacky309, line 89: no need for a signature for local IPC

src/ipc/local/LocalIPCRequestHandler.h
```diff
@@ -61,9 +60,10 @@ class FaceliftIPCLocalLib_EXPORT LocalIPCRequestHandler
 {
 
 public:
-    virtual ~LocalIPCRequestHandler() = default;
-    virtual void deserializePropertyValues(LocalIPCMessage &msg, bool isCompleteSnapshot) = 0;
-    virtual void deserializeSignal(LocalIPCMessage &msg) = 0;
+    virtual void unmarshalProperties(const QVariantMap& changedProperties) = 0;
```
jacky309: no QVariant in local IPC

src/ipc/local/LocalIPCServiceAdapter.h
roam3r: The LocalIPCServiceAdapterBase.h has method "appendDBUSIntrospectionData" which was probabely introduced to treat the ServiceAdapter layer consistenly so is "typeToSignature".

codegen/facelift/templates/Struct.template.cpp
roam3r: The "treatAsByteArray" is a structure annotation, adding that to the structure properties is only natural.
Moving it out needs introduction of a whole new layer for DBusIPC to be able to treat structs generically with no benefit what-so-ever.
Sorry I won't do that, those kind over-engineered decisions (like having all structs internal properties exposed as tuple for some other layer to be able to serialize it!!!) has made facelift a highly unnecessary complicated lib.

src/ipc/local/LocalIPCRequestHandler.h
roam3r: So let's look at what QVariant replaces. QVariant replaces the "facelift serialization" mechanism which in first place was reinvention of the wheel. You don't need to serialize/deserialize properties if you need only to treat them in the same process generically. 
Anyhow in terms of difference now we are looking at cost of serializing a strut (facelift serialization) vs cost of cloning it (QVariant) and if the latter deems to be more costly then the former it again can be traced back to wrong fundamentals of facelift::structure class and overhead of initialization std::tuples.  
So "no QVariant in local IPC" is not a technical review comment.
If you want to do facelift any good I suggest we rethink the facelift::structure.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -134,7 +187,7 @@ void {{className}}::deserializeSignal(InputIPCMessage &msg)
 void {{className}}::set{{property}}({{property.cppMethodArgumentType}} newValue)
 {
     {% if (not property.type.is_interface) %}
-    ipc()->sendSetterCall(memberID(MethodID::set{{property.name}}, "set{{property.name}}"), newValue);
+    ipc()->sendSetterCall("{{property.name}}", newValue);
```
jacky309: FYI, the construct you broke here is meant to provide a way for that generated code to be usable with both an IPC transport channel where method IDs are strings (such as DBus), and an IPC where method IDs are integer (such as the the current "local" IPC, and the hi-performance IPC which I have been working on). Using integers instead of strings is obviously more efficient.
Your change will make the introduction of a new IPC harder.

src/ipc/local/LocalIPCMessage.cpp
roam3r: You need to create facelift specific benchmarks, I created ones with function calls with different range of structs and size.
Please see my other comment about QVariant and facelift payload.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: Can we please stop the "performance" discussion? now you are discussing about "future possible performance degrades with some not in-place high performance IPC!!!".

idleroamer:
@jacky309 Please let me know if you have any further comments, I assume the current ones are all addressed.

src/ipc/ipc-common/CMakeLists.txt
```diff
@@ -1,10 +1,9 @@
+find_package(Qt5DBus REQUIRED)
```
jacky309: QtDBus is not available on android.

jacky309 (changes requested):

src/ipc/ipc-common/CMakeLists.txt
roam3r: thanks for spotting

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
...
-    } else
+    }
     {% endfor %}
+
+    {% if interface.hasModelProperty %}
+    this->onModelUpdateEvent(msg);
+    {% endif %}
+}
+
+const QList<QString>& {{className}}::getSignals() const
```
jacky309, line 115: this is relevant to the DBus layer only.

codegen/facelift/templates/IPCServiceAdapter.template.h
```diff
@@ -83,7 +82,13 @@ class {{classExport}} {{className}}: public {{baseClass}}
 
     void connectSignals() override;
 
-    void serializePropertyValues(OutputIPCMessage& msg, bool isCompleteSnapshot) override;
+    QVariantMap changedProperties();
```
jacky309: changedProperties vs marshalProperties ??

codegen/facelift/templates/Struct.template.cpp
jacky309: make this a DBus-specific trait of the structure.

src/ipc/dbus/FaceliftDBusMarshaller.h
```diff
@@ -0,0 +1,278 @@
...
+    int tmp;
+    argument >> tmp;
+    source = static_cast<T>(tmp);
+    return argument;
+}
+
+template<typename T, typename std::enable_if_t<std::is_convertible<T, facelift::StructureBase>::value, int> = 0>
+inline QDBusArgument &operator<<(QDBusArgument &argument, const T &structure)
+{
+    if (!treatAsByteArray(structure)) {
```
jacky309: turn this into a compile time decision.

src/ipc/dbus/FaceliftDBusMarshaller.h
```diff
@@ -0,0 +1,278 @@
...
+        argument << byteArray;
+        argument.endStructure();
+        return argument;
+    }
+}
+
+template<typename T, typename std::enable_if_t<std::is_convertible<T, facelift::StructureBase>::value, int> = 0>
+inline const QDBusArgument &operator>>(const QDBusArgument &argument, T &structure)
+{
+    if (!treatAsByteArray(structure)) {
```
jacky309: compile time decision

src/ipc/dbus/IPCDBusServiceAdapterBase.h
```diff
@@ -37,11 +37,15 @@
 #endif
 
 #include <QDBusVirtualObject>
+#include <QtDBus>
+#include <mutex>
```
jacky309: you include that header several times but you do not seem to use it

doc/page-annotations.h
```diff
@@ -39,6 +39,7 @@ The following annotations can be used in structure definitions:
 
 |Annotation                     |              Description                                    | Supported over IPC |
 |-------------------------------|-------------------------------------------------------------|--------------------|
+|\@treatAsByteArray: true       | Treat the whole structure as byte array over IPC reducing structure signature only to DBUS Type ARRAY OF BYTE | Yes             |
```
jacky309: the annotation name should contain "ipc", or "dbus". Something like:
@dbus-serialization: byteArray

codegen/facelift/templates/IPCServiceAdapter.template.h
roam3r: what do you mean? these two functions have completely different purposes.

doc/page-annotations.h
roam3r: i think we only need a true/false value, the "byteArray" as a value suggests there are other possibilities as well.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: I consider moving the function to anther layer over-complicating the design with no apparent benefits. After all this adds no overhead when not used in "local"

idleroamer:
@jacky309 Can you please let me know what is still blocking the merge?

jacky309:
> @jacky309 Can you please let me know what is still blocking the merge?

Just check my latest comments.

codegen/facelift/templates/IPCServiceAdapter.template.cpp
```diff
@@ -109,42 +100,69 @@ void {{className}}::appendDBUSIntrospectionData(QTextStream &s) const
 {
     Q_UNUSED(s);   // For empty interfaces
     {% for property in interface.properties %}
-    ::facelift::DBusSignatureHelper::addPropertySignature<ServiceType::PropertyType_{{property.name}}>(s, "{{property.name}}", {{ property.readonly | cppBool }});
+    {% if not property.type.is_model %}
```
jacky309, line 104: The existing approach based on DBusSignatureHelper ensures that all properties of a given type share the same code, which results in a more compact binary code. This is not the case with your change.

codegen/facelift/templates/IPCServiceAdapter.template.cpp
```diff
@@ -109,42 +100,69 @@ void {{className}}::appendDBUSIntrospectionData(QTextStream &s) const
...
+        s << "\" direction=\"out\"/>";
         {%- endfor -%}
-        >(s, "{{signal.name}}", argumentNames);
+        s << "</signal>";
     }
-
     {% endfor %}
+
+    {% if interface.hasModelProperty %}
+    s << "<signal name=\"" << facelift::IPCCommon::MODEL_DATA_CHANGED_MESSAGE_NAME << "\">";
```
jacky309: that code is generic so it should not be duplicated.

codegen/facelift/templates/IPCServiceAdapter.template.cpp
roam3r: The existing approach based on DBusSignatureHelper was **simply wrong**, go ahead and check.
So a change is needed to fix a defect!

idleroamer:
> > @jacky309 Can you please let me know what is still blocking the merge?
> 
> Check the comments I made during the last days.

All of them are addressed.

codegen/facelift/templates/IPCServiceAdapter.template.cpp
roam3r: The whole "model" should be treated as separate interface (similar to org.freedesktop.DBus.Properties with its own signals and methods) so these will be improved automatically.

codegen/facelift/templates/IPCServiceAdapter.template.cpp
jacky309: A change is needed indeed, but the approach can be kept.

jacky309:
> > > @jacky309 Can you please let me know what is still blocking the merge?
> > 
> > 
> > Check the comments I made during the last days.
> 
> All of them are addressed.

Rejecting is not addressing.

idleroamer:
> > > > @jacky309 Can you please let me know what is still blocking the merge?
> > > 
> > > 
> > > Check the comments I made during the last days.
> > 
> > 
> > All of them are addressed.
> 
> Rejecting is not addressing.

Either you are having a dialogue about the review points which involves you delivering arguments against the latest comments or otherwise they are consider resolved. Or you consider this a one-way street which is not a open source culture and there is no point for either of us continue this.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
jacky309: you can probably alter the code generator to produce DBus-related methods only when generating DBus classes.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
jacky309: The local IPC uses integer IDs, so we are not talking about the future.. The current approach should work with your changes.

src/ipc/local/LocalIPCRequestHandler.h
jacky309: I do not understand what the point about std::tuples has to do with that discussion. Even if we were using "plain" data members instead of packing them into a tuple, we would not like to convert anything to QVariant or QVariantMap.
I expect the current serialization to be significantly faster than packing the data into a QVariantMap.

roam3r: The structs need to be treated generically in "local" mode.
The QVariant replaces the LocalIPCMessage payload, right?
In other words cloning (QVariant) replaces serialization (LocalIPCMessage payload).
And why on earth cloning an object should be slower than serializing it?
If that is the case in facelift, the only reason is the overhead of creating std::tuples when cloning a struct.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: The performance claims needs to be backed up by actuall KPI measurement
There "Local" version AFAIK uses internal Qt signal/slots so we can't have seriously a discussion over difference of treating a signal with integer and a signal with string!

jacky309: I have no doubt that sending an integer (which does not even need to be wider than 8 bits) is faster than sending a multi-character string. Also, string comparisons are not free, so the processing of an incoming request is also faster if an integer is used as method ID since no comparison is even needed.

src/ipc/local/LocalIPCRequestHandler.h
jacky309: > The structs need to be treated generically in "local" mode.
> The QVariant replaces the LocalIPCMessage payload, right?

The QVariantMap does.

> In other words cloning (QVariant) replaces serialization (LocalIPCMessage payload).

The creation and population of the QVariantMap is what needs to be compared against the serialization.
We are talking about a map where the keys are QStrings and the elements are QVariants. The insertion of one element in that map involves the hashing of the key string, the conversion of the value to QVariant, ....
On the receiver side, a lookup in the map is needed, some type checking, etc....
In comparison, serializing/deserializing the value to a QByteArray is fast, unless the value is a large container which supports implicit sharing. Do you have such containers in mind ?

> And why on earth cloning an object should be slower than serializing it?
> If that is the case in facelift, the only reason is the overhead of creating std::tuples when cloning a struct.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: A simple benchmark of signal 30 character long being emitted/received 10000 times shows followings:
https://github.com/idleroamer/facelift/tree/feature/benchmark_local_signals

Current implementation
Signal emission/receive duration: 1463 ms

Current PR: 
Signal emission/receive duration:1102 ms

A signal name comparison is not a bottleneck.

src/ipc/local/LocalIPCRequestHandler.h
roam3r: All of the operations in creating a QVariantMap is negligible compared to creation of QVariant itself (backed by KPI) 
unless you're interface has over tens of thousands of properties inside!!!

Premature optimization is the root of all evil.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: yes, could be but there are other examples (like signatureHelper) inside Adapter layer which actually belong to dbus layer, I suggest to create an issue for that so that all of them can be addressed together.

codegen/facelift/templates/Struct.template.cpp
roam3r: and how do you suggest to do that?

codegen/facelift/templates/IPCProxyAdapter.template.cpp
jacky309: Do those figures tell us that using a string is faster than using an integer ?
The reason why integers can be used is not really because using strings has been identified as a performance hotspot, but rather that it is necessary for IPCs which use integers as messageID. I do not understand why your change has to break that.

roam3r: No, those figures tell us whether we send MessageIDs as strings or as integers is not a performance bottleneck.
It as well tells us performance arguments without benchmarks has no values.

I appreciate your time but I have the feeling you even haven't looked at the PR? the "sendSetterCall" is internal call between IPCProxy and Local/DBusIPCProxyBinder and using integers is simply superfluous

You had performance claims now you are talking again about some IPC which is not even there yet AGAIN.
Please let me know if you personally hold this PR against your views/plan for facelift so I don't take your time anymore.

src/ipc/local/LocalIPCRequestHandler.h
jacky309: I agree that premature optimization is evil, and the current code is not the result of a premature optimization. Using the same serialization code in the local IPC as we do in the DBus IPC was a way to easily and quickly enable the local IPC. If high performance had really been required, no serialization would have been used.

I don't really see any nice alternative to the usage of a QVariant here though. Using std::variant could be better since it does not require all the type registrations and overhead which QVariant requires, but it would probably result in a more complex code.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
jacky309: There is a misunderstanding here.
How about having a call tomorrow afternoon ? I guess we could use google meet for that.

roam3r: That is a fine suggestion. Though tomorrow is a little tight in the afternoon, so let me know if tomorrow morning or the Friday's afternoon fits your calendar, otherwise we postpone to next week.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
...
-    {% for property in interface.properties %}
-    {% if property.type.is_interface %}
-    bool emit_{{property.name}}ChangeSignal = false;
-    QString {{property.name}}_objectPath;
-    if (deserializeOptionalValue(msg, {{property.name}}_objectPath, isCompleteSnapshot))
-    {
-        m_{{property.name}}Proxy.update({{property.name}}_objectPath);
-        m_{{property.name}} = m_{{property.name}}Proxy.getValue();
-        emit_{{property.name}}ChangeSignal = true;
+    QMap<QString, bool> emitChangeSignal;
```
jacky309: There is no need to use a QMap here. keep the existing approach with one local bool variable per property.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
...
-        deserializeValue(msg, {{property.name}}Size);
-        m_{{property.name}}.beginResetModel();
-        m_{{property.name}}.reset({{property.name}}Size, std::bind(&ThisType::{{property.name}}Data, this, std::placeholders::_1));
-        m_{{property.name}}.endResetModel();
-        emit_{{property.name}}ChangeSignal = true;
+    for (const QString &propertyName: emitChangeSignal.keys()) {
+        {% for property in interface.properties %}
+        if (propertyName == QStringLiteral("{{property.name}}") && emitChangeSignal[propertyName]) {
+            emit {{property.name}}Changed();
+        }
```
jacky309, line 87: } else

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
...
+            int {{property.name}}Size = castFromQVariant<int>(values[propertyName]);
+            m_{{property.name}}.beginResetModel();
+            m_{{property.name}}.reset({{property.name}}Size, std::bind(&ThisType::{{property.name}}Data, this, std::placeholders::_1));
+            m_{{property.name}}.endResetModel();
+        {% else %}
+            const auto previous_{{property.name}}_Value = m_{{property.name}};
+            m_{{property.name}} = castFromQVariant<{{property.interfaceCppType}}>(values[propertyName]);
+            emitChangeSignal[QStringLiteral("{{property.name}}")] = ((previous_{{property.name}}_Value != m_{{property.name}}));
+        {% endif %}
+        }
```
jacky309: } else

codegen/facelift/templates/IPCProxyAdapter.template.cpp
jacky309: When a proxy successfully connects to its server counterpart, some application code may legitimately assume that the "readyChanged" signal is triggered after all other property change signals, since "readyChanged with ready==true" means that the proxy is fully initialised and ready to be used. Make sure this order is preserved.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
...
         {{parameter.interfaceCppType}} param_{{parameter.name}};
-        deserializeValue(msg, param_{{parameter.name}});
+        param_{{parameter.name}} = (argumentsIterator.hasNext() ? castFromQVariant<{{parameter.interfaceCppType}}>(argumentsIterator.next()):{% if not parameter.type.is_interface %}{{parameter.interfaceCppType}}(){% else %}nullptr{% endif %});
         {% endfor %}
         emit {{event}}(
         {%- set comma = joiner(", ") -%}
         {%- for parameter in event.parameters -%}
             {{ comma() }}param_{{parameter.name}}
         {%- endfor -%}  );
-    } else
```
jacky309, line 116: keep that "else"

codegen/facelift/templates/IPCProxyAdapter.template.cpp
```diff
@@ -55,76 +51,132 @@
     {% endif %}
 }
 
-void {{className}}::deserializePropertyValues(InputIPCMessage &msg, bool isCompleteSnapshot)
+void {{className}}::unmarshalProperties(const QVariantMap& values)
```
jacky309, line 54: AFAIS, unmarshalProperties() and unmarshalPropertiesChanged() have a lot in common so they can use the same code path, just like it is in the original code. In case it was not clear enough, the "isCompleteSnapshot" bool argument is used to differentiate the initial property transfer (where the complete set of property values is exchanged) from a later transfer, where only the changed values are exchanged.

roam3r: Well logically they are different (even more now due to last change based on your comment), the input and inner algorithm are really different. So this is not a logic duplication which would justify merging those together.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: good point.

codegen/facelift/templates/IPCProxyAdapter.template.cpp
roam3r: The logic is different now, there is no logical meaning to a "else" branch anymore.

idleroamer:
Please remove "request a change", no points are open.

jacky309:
You mentioned a noticeable drop in performance. Could you please provide some details ?

Most people simply do not care about having a "native" DBus serialization but they do care about having an IPC layer which performs well with their use cases. I do not think it is reasonable to ask anyone to work around a performance drop introduced by a feature which they do not need.

idleroamer:
I took a very extreme example where a IPC interface has methods with structs over hundreds of fields, making RPC calls hundreds of times in a row, to see how libdbus acts. 
In real use-cases with moderate size structs and reasonable num of IPC calls, the overhead is not even measurable.

So "Nobody" won't feel any difference! We have a huge project with all possible use-cases and observed no difference.

Anyhow if for any reason such unrealistic use-case (hunders of fields in struct) is needed, "toByteArrayOverDBus" annotation for structs will relieve one from signature validation in libdbus and remove any possible overhead.

The annotation should be on struct level as they could cause overhead with their size not the interface.

jacky309:
> I took a very extreme example where a IPC interface has methods with structs over hundreds of fields, making RPC calls hundreds of times in a row, to see how libdbus acts.
> In real use-cases with moderate size structs and reasonable num of IPC calls, the overhead is not even measurable.
> 
> So "Nobody" won't feel any difference! We have a huge project with all possible use-cases and observed no difference.

Unless this has changed recently, your project is mostly single-process, so I understand that the difference can hardly be noticed... There was a noticeable performance improvement when switching to the QByteArray serialization when multi-process was used.

> 
> Anyhow if for any reason such unrealistic use-case (hunders of fields in struct) is needed, "toByteArrayOverDBus" annotation for structs will relieve one from signature validation in libdbus and remove any possible overhead.

Lists of reasonably large structs should be supported with reasonable performance. That kind of list can easily contain up to hundred fields.

> 
> The annotation should be on struct level as they could cause overhead with their size not the interface.

Either an interface requires the interoperability feature and it needs to be carefully designed to perform well over IPC, or it does not use need interoperability (which is probably the most common situation) and it should benefit from an optimal serialization performance and not have any limitation in terms of signature length. The decision is based on how the interface is used, so I strongly believe that the annotation should belong to the interface definition.
Additional reasons:
     - Data types can be used in multiple interfaces. Enabling the "toByteArrayOverDBus" flag on a type could affect some interfaces which need the interoperability features, making them unusable (or partly usable).
     - Whether the signature length limit is reached or not depends on what types are packed together in a single DBus message. It could be that an interface using a large struct does not reach the limit, while another interface using the same struct hits the limit because it is used in combination with other large types.

idleroamer:
Obviously the Multi-process part is in focus. My benchmarks don't show any measurable performance change in realistic cases, we are talking about less than fractions of ms per calls.
I can't believe the improvement you observed by switching to QByteArray was relevant to scope of this change.
> 
> Unless this has changed recently, your project is mostly single-process, so I understand that the difference can hardly be noticed... There was a noticeable performance improvement when switching to the QByteArray serialization when multi-process was used.
> 
Lists/Maps doesn't increase the libdbus overhead, signature is increased only by "a()". 
Obviously more data is transferred in case of lists but the cost is constant in both current implementation and current PR.
i.e. for each additional entry in the list it takes the same exact amount of time longer.
> 
> Lists of reasonably large structs should be supported with reasonable performance. That kind of list can easily contain up to hundred fields.
> 
It is not a feature to be conformant with the DBus rules, it is a bug not to be.
The question is what is the default behavior rather than where annotation should go.
My analysis shows the serialization to ByteArray over DBus has very very small advantage in real use-cases so that it should be used for very corner cases (if at all required). We should encourage developers to design the interfaces more modular rather than using huge structs.
I see pros/cons in both.
> 
> Either an interface requires the interoperability feature and it needs to be carefully designed to perform well over IPC, or it does not use need interoperability (which is probably the most common situation) and it should benefit from an optimal serialization performance and not have any limitation in terms of signature length. The decision is based on how the interface is used, so I strongly believe that the annotation should belong to the interface definition.
> Additional reasons:
> - Data types can be used in multiple interfaces. Enabling the "toByteArrayOverDBus" flag on a type could affect some interfaces which need the interoperability features, making them unusable (or partly usable).
No it is not how it works, the limit is per method argument not the accumulation of them.
> - Whether the signature length limit is reached or not depends on what types are packed together in a single DBus message, which means it depends on the interface definition. It could be that an interface using a large struct does not reach the limit, while another interface using the same struct hits the limit because it is used in combination with other large types (think about multiple method arguments, or multiple properties).

PR #307, opened 2020-09-22
Add auxiliary API for IPCProxy

pfagner:


src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -101,6 +100,30 @@ class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
...
+                                           context, [this, f{std::move(f)}, connection{std::move(connection)} ]()
+                {
+                    if (this->ready()) {
+                        f();
+                        QObject::disconnect(*connection);
+                    }
+                }
+            );
+        }
+    }
```
bitmouse, line 126: The whole idea was to NOT add it this way but to create an external observer. This causes this class to grow.

bitmouse (changes requested):
The whole idea was to NOT add it this way but to create an external observer. This causes this class to grow.

src/ipc/ipc-common/IPCProxyBase.h
jacky309: Such a helper is useful for any class which extends InterfaceBase, not only for IPCProxies.

bitmouse: > Such a helper is useful for any class which extends InterfaceBase, not only for IPCProxies.

exactly. That is why it should be added as external observer class, not inside just one class.

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+// IsReadyObserver
+class IsReadyObserver : public QObject
+{
+    Q_OBJECT
+    bool m_ready{};
+public:
+    IsReadyObserver(QObject *parent = 0) : QObject(parent) {};
+    Q_SIGNAL void isReadyChanged();
+
+    bool ready(){ return m_ready; }
```
bitmouse, line 89: const

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    bool set_ready(bool val){
+        return m_ready=val;
+    }
```
bitmouse, line 92: formatting

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    {
+        this->m_ready = val;
+        notify(*this);
+    }
+};
+// IsReadyObserver
+class IsReadyObserver : public QObject
+{
+    Q_OBJECT
+    bool m_ready{};
```
bitmouse, line 84: {} is not needed here, however it would be nice to have it initialized explicitly as false

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    bool ready(){ return m_ready; }
+    bool set_ready(bool val){
+        return m_ready=val;
+    }
+    void raiseSignal(){
+        emit isReadyChanged();
+    }
+
+    template<typename Function>
+    void callReady(const QObject* context, Function f)
```
bitmouse, line 98: onReady

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+class IsReadyObserver : public QObject
+{
+    Q_OBJECT
+    bool m_ready{};
+public:
+    IsReadyObserver(QObject *parent = 0) : QObject(parent) {};
+    Q_SIGNAL void isReadyChanged();
+
+    bool ready(){ return m_ready; }
+    bool set_ready(bool val){
```
bitmouse, line 90: please keep camelCase style. Also 'val' is not the nicest name, I would suggest 'ready'

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+
+    bool stateReady() const { return m_ready; }
+    void stateReadyChanged(bool val)
+    {
+        this->m_ready = val;
+        notify(*this);
+    }
+};
+// IsReadyObserver
+class IsReadyObserver : public QObject
```
bitmouse, line 81: what is the observed object?

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    IsReadyObserver(QObject *parent = 0) : QObject(parent) {};
+    Q_SIGNAL void isReadyChanged();
+
+    bool ready(){ return m_ready; }
+    bool set_ready(bool val){
+        return m_ready=val;
+    }
+    void raiseSignal(){
+        emit isReadyChanged();
+    }
```
bitmouse, line 95: not needed

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+                    }
+                }
+            );
+        }
+    }
+};
+// Single-time observer which will unregister itself when done
+class SingleTimeObserver : public IsReadyObserver, public IObserver
+{
+    const QObject* m_context{};
```
bitmouse, line 120: remove {}. You initialize it in constructor!!!!!

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    std::function<void()> m_func{};
```
bitmouse, line 121: same as above

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+public:
+    explicit SingleTimeObserver(const QObject* context, std::function<void()> func=nullptr): m_context{context}, m_func{func} {}
+    ~SingleTimeObserver()=default;
+
+    void handle_event(Monitor &source ) override {
+        if(set_ready(source.stateReady())){
+            callReady(m_context, m_func);
+            raiseSignal();
+            source.del(*this);
+        }
```
bitmouse, line 131: I don't get it. How it should work?

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
...
+    void handle_event(Monitor &source ) override {
+        if(set_ready(source.stateReady())){
+            callReady(m_context, m_func);
+            raiseSignal();
+        }
+    }
+};
 
+template<typename AdapterType>
+class IPCProxyBase : public AdapterType, Monitor, protected IPCProxyBaseBase
```
bitmouse, line 152: You should not extend class further. Inheritance is the class extension. Please create member object that holds observers.

bitmouse (changes requested):

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -43,23 +42,132 @@
 
 namespace facelift {
 
-template<typename AdapterType>
-class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
+class Monitor;
+class IObserver
```
jacky309, line 46: What does this code enable, which the code from the very first commit of that PR did not ?
This commit looks horribly complex. Why not simply moving the code from the first commit into a helper class ?

pfagner: Please don't worry, I take account all comments/notes.
All code will be rewritten.

pfagner:
By reason broken fork that PR has been broken.

PR #308, opened 2020-09-22
Add HEADERS_NO_MOC, HEADER_NO_NOC_FILE options for excluding some files from processing by moc

kunichik:


cmake/faceliftMacros.cmake
```diff
@@ -405,8 +407,13 @@ macro(_facelift_add_target_start IMPLEMENTATION_TARGET_NAME)
     set(HEADERS_TO_BE_MOCCED ${HEADERS} ${HEADERS_NO_INSTALL})
     unset(HEADERS_MOCS)
 
+    # exclude unnecessary files mocking
```
bitmouse, line 410: this is not mocking!
e.g. "exclude unnecessary files from running moc"

cmake/faceliftMacros.cmake
```diff
@@ -405,8 +407,13 @@ macro(_facelift_add_target_start IMPLEMENTATION_TARGET_NAME)
...
+    foreach(HEADER ${HEADERS_NO_MOC})
+        string(REPLACE "${HEADER}" "" HEADERS_TO_BE_MOCCED "${HEADERS_TO_BE_MOCCED}")
```
bitmouse, line 412: HEADERS_TO_BE_MOCCED sounds weird :) If you could find better name it would be beneficial (otherwise keep it as is)

src/common/CMakeLists.txt
```diff
@@ -4,5 +4,8 @@ facelift_add_library(FaceliftCommonLib
     HEADERS
         span.h
         FaceliftUtils.h
+    HEADERS_NO_MOC
```
jacky309, line 7: How does HEADERS_NO_INSTALL work with HEADERS_NO_MOC ?

cmake/faceliftMacros.cmake
kunichik: agree, I have noticed that this word doesn't describe enough the purpose. Will be changed.

cmake/faceliftMacros.cmake
kunichik: it's the existing variable, I try to change as little as possible, because this part is very dangerous.
Will be changed though.

src/common/CMakeLists.txt
kunichik: Nohow, I'll take care of it.

kunichik: But no, it is works like this:
HEADERS_NO_MOC are excluded from both HEADERS and HEADERS_NO_INSTALL.
Is this the correct way ?

bitmouse: IMO this is correct way. If we do not want to run moc and don't want to install it, we do not need to add the header at all.

jacky309: we need to provide a way to install headers which do not need the moc.

HEADERS_NO_INSTALL adds new headers (whitelist) whereas your implementation of HEADERS_NO_MOC removes (blacklist) content from list of headers. This is inconsistent.

bitmouse: Actually HEADERS_NO_MOC should install headers. @kunichik please test if it works this way.
Summary:
- HEADERS -> run moc, install
- HEADERS_NO_INSTALL -> run moc, don't install
- HEADERS_NO_MOC -> don't run moc, install
- header not added -> don't run moc, don't install (no need to add header like that)

kunichik: yep, I've not added that
HEADERS_NO_MOC are excluded from both HEADERS and HEADERS_NO_INSTALL and write to the new variable HEADERS_FOR_MOC. The variable sends as argument for moc.
HEADERS and HEADERS_NO_INSTALL themself don't change at all.

> HEADERS -> run moc, install
true
> HEADERS_NO_INSTALL -> run moc,
true, but
>  don't install
Except of adding to moc processing, HEADERS_NO_INSTALL don't use anywhere.
In the previous realization HEADERS_NO_INSTALL had been only added to the moc processing:
`set(HEADERS_FOR_MOC ${HEADERS} ${HEADERS_NO_INSTALL})`
 that's it
    
> HEADERS_NO_MOC -> don't run moc, install
true for those headers, that are in HEADERS.
HEADERS_NO_MOC is only removes headers from HEADERS and HEADERS_NO_INSTALL

>  header not added -> don't run moc, don't install (no need to add header like that)
true

cmake/faceliftMacros.cmake
```diff
@@ -388,6 +390,26 @@ macro(_facelift_parse_target_arguments additionalOptions additionalOneValueArgs
         list(APPEND HEADERS ${GLOB_FILES})
     endforeach()
 
+    set(HEADERS_NO_MOC ${ARGUMENT_HEADERS_NO_MOC})
+
+    # files for installation can be in both sections: HEADERS and HEADERS_NO_MOC
```
jacky309, line 395: I see no reason why having a header in both sections should be allowed.
We should define one proper way of doing things.

doc/page-cmakefunctions.h
```diff
@@ -10,6 +10,9 @@ The following functions are public and are meant to be used by FaceLift users.
 facelift_add_interface(<TargetName>
                        INTERFACE_DEFINITION_FOLDER dir
                        [LINK_LIBRARIES lib1 [lib2...]]
+                       [HEADERS_NO_MOC hdr1 [hdrnm2...]]
```
jacky309, line 13: there is no header argument to be added here

doc/page-cmakefunctions.h
```diff
@@ -46,6 +49,9 @@ facelift_add_qml_plugin(<TargetName>
                         [HEADERS_GLOB [Globbing expressions]...]
                         [SOURCES_GLOB_RECURSE [Globbing expressions]...]
                         [HEADERS_GLOB_RECURSE [Globbing expressions]...]
+                        [HEADERS_NO_MOC hdr1 [hdrnm2...]]
```
jacky309, line 52: hdrnm2 ??

examples/AddressBook/CMakeLists.txt
```diff
@@ -1,6 +1,11 @@
 
 facelift_add_interface(AddressBookInterface
     INTERFACE_DEFINITION_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/interface
+    HEADERS_GLOB_RECURSE_NO_MOC
```
jacky309, line 4: users of facelift_add_interface do not know which files are generated. They know even less about which files need to be MOCed.

src/model/CMakeLists.txt
```diff
@@ -78,6 +78,54 @@ facelift_add_library(FaceliftModelLib
         StructTypeHandler.h
         PropertyInterface.h
         FaceliftProperty.h
+    HEADERS_NO_MOC
+        ModelProperty.h
```
jacky309, line 82: Each header should be defined only once.

cmake/faceliftMacros.cmake
kunichik: changed, added a check.
let me know if there is not too much inform messages
I've added them only to make the error clear to the user

src/model/CMakeLists.txt
kunichik: fixed

examples/AddressBook/CMakeLists.txt
kunichik: It seems that the same files are generating the warnings
* Module.h
* ModuleIPC.h
* ModulePrivate.h
* ObjectRegistryIPCCommon.h
* {interface_name}IPCCommon.h
     e.g. AddressBookIPCCommon.h and SubInterfaceIPCCommon.h

How are you see if we'll create some moc blacklist for such files 
instead of adding them to each CMakeLists.txt ?

jacky309: you can let the code generator generate a new file containing the following:
_set(GENERATED_HEADERS_MOC
    File1.h
    File2.h
    ....
)
set(GENERATED_HEADERS_NO_MOC
    File3.h
    File4.h
    ....
)
set(GENERATED_SOURCES
    File1.cpp
    File2.cpp
    ....
)_

In order to identify which files should be MOCed, the code generator can search for "Q_OBJECT" in the files.
Then, the implementation of the "facelift_add_interface" function can use the "include" command (https://cmake.org/cmake/help/v3.0/command/include.html) to get the definitions of those variables and call the facelift_add_library with those file lists.

bitmouse: I 100% agree with the first part, about generating config files. We know, during the code generation, which files require running moc. 

However searching for Q_OBJECT is most probably wrong. We would have to look also for Q_GADGET and probably a few other macros. Someone could also define Q_OBJECT inside other macro (not applicable to generated code) and we may miss it. Another point: this is slower for generated code, for which we already know if we have to run moc. I would suggest to keep it to developer decision. 

This way or another, user should not add generated headers to HEADERS_NO_MOC. The user should add only his headers. This applies to all macros.

kunichik: one problem that our templates can generate or not Q_OBJECT macro.
e.g. build/facelift_generated/MyPackageInterface/module/facelift/example/mypackage/ModulePrivate.h doesn't have Q_OBJECT
but 
build/facelift_generated/MyPackageInterface/module/facelift/example/mypackage/ModulePrivate.h
does not
because codegen/facelift/templates/ModulePrivate.template.h 
has for-loop in which we can omit Q_OBJECT generating.

so probably Generator must search for some of macros, 
but those macros could be defined in CMAKE_AUTOMOC_MACRO_NAMES (https://cmake.org/cmake/help/v3.12/variable/CMAKE_AUTOMOC_MACRO_NAMES.html#variable:CMAKE_AUTOMOC_MACRO_NAMES)

that's why we should search for Q_OBJECT etc. via cmake not in Generator

kunichik: I've just forgot that we do not use AUTOMOC...

so we can define CMAKE_MOC_MACRO_NAMES as follows
   
set(CMAKE_MOC_MACRO_NAMES "Q_OBJECT" "Q_GADGET" "Q_NAMESPACE")

btw. why do not we use AUTOMOC?

kunichik:
I've dune some timings.
I've just run current master's version and the one from this PR one by one.
It seems that we have achieved some speed up:

Facelift without extra moc processing | Current Github version
-- | --
1:27 | 1:48
1:03 | 1:05
1:20 | 1:47
1:30 | 1:08
1:10 | 1:56
0:49 | 1:07
1:03 | 1:23
1:17 | 1:33
0:52 | 1:01

we achieve less then 1 minute  time(last four rows) after today's last push (with HEADERS_CAN_MOC_WILDCARD, HEADER_NO_MOC_WILDCARD and deleting one more HEADERS cycle before moc processing  )

cmake/faceliftMacros.cmake
```diff
@@ -388,6 +390,30 @@ macro(_facelift_parse_target_arguments additionalOptions additionalOneValueArgs
...
+        string(FIND ";${HEADERS_NO_MOC}" ";${HEADER}" IS_FOUND )
+        if(NOT ${IS_FOUND} LESS 0)
+            message("HEADERS_NO_MOC is '${HEADERS_NO_MOC}'")
+            message("HEADERS is '${HEADERS}'")
+            message("HEADER is '${HEADER}'")
+            message(FATAL_ERROR "Header '${HEADER}' is present in both HEADERS and HEADERS_NO_MOC sections at position '${IS_FOUND}'")
+        endif()
+    endforeach()
+
+    UNSET(HEADERS_NO_MOC_GLOB)
```
bitmouse, line 406: HEADERS_GLOB_NO_MOC vs HEADERS_NO_MOC_GLOB? It is confusing. Please improve naming.

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+
+                if(${IS_FOUND} GREATER_EQUAL 0)
+                    break()
+                endif()
+
+            endforeach()
+
+            if(${IS_FOUND} LESS 0)
+                list(APPEND HEADERS_NO_MOC_GLOB ${HEADER_CAN_HAVE_MOC})
+            endif()
```
bitmouse, line 476: What if we find the macro?

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+
+        endforeach()
+    endif()
+
+    # we prepend string like ";header1.h;header2.h;" and then replace 
+    # like ;header1.h just not to replace header like extraheader1.h
```
bitmouse, line 482: I think this comment is not relevant anymore

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+    set(HEADERS_FOR_MOC ${HEADERS} ${HEADERS_NO_INSTALL})
+
     unset(HEADERS_MOCS)
 
-    if(HEADERS_TO_BE_MOCCED)
-        qt5_wrap_cpp(HEADERS_MOCS ${HEADERS} ${HEADERS_NO_INSTALL} TARGET ${IMPLEMENTATION_TARGET_NAME})
+    # exclude unnecessary auto generated or set by wildcard files from running moc
+    foreach(HEADER ${HEADERS_NO_MOC_GLOB})
+        string(REPLACE ";${HEADER}" "" HEADERS_FOR_MOC ";${HEADERS_FOR_MOC}")
```
bitmouse, line 489: Are there any list based solutions? It would be better than direct string manipulation

examples/MyPackage/CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 
 facelift_add_interface(MyPackageInterface
     INTERFACE_DEFINITION_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/interface
+    # TODO: describe such parameter in the documentation
+    #HEADERS_GLOB_RECURSE_NO_MOC 
+    #    ${CMAKE_BINARY_DIR}/facelift_generated/*ModuleIPC.h
```
bitmouse, line 6: remove

examples/MyPackage/CMakeLists.txt
```diff
@@ -9,6 +12,8 @@ facelift_add_qml_plugin(MyPackagePlugin
     URI facelift.example.mypackage
     SOURCES_GLOB_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp
     HEADERS_GLOB_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/*.h
+    HEADERS_GLOB_RECURSE_NO_MOC 
+        ${CMAKE_CURRENT_SOURCE_DIR}/models/cpp/MyInterfaceCppImplementation.h
```
bitmouse, line 16: this is not GLOB

examples/Tuner/CMakeLists.txt
```diff
@@ -9,6 +9,10 @@ facelift_add_qml_plugin(TunerModelPlugin
     URI tuner
     SOURCES_GLOB_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp
     HEADERS_GLOB_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/*.h
+    # try to choose NO MOC path as local as possible
+    # in this case we use the files only for TunerInterface:
+    # HEADERS_GLOB_RECURSE_NO_MOC ${CMAKE_BINARY_DIR}/facelift_generated/TunerInterface/*Module.h
```
bitmouse, line 14: remove

bitmouse (changes requested):

examples/Tuner/CMakeLists.txt
kunichik: done

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+       ${CURRENT_GENERATED_DIR}/*/ModuleIPC.h
+       ${CURRENT_GENERATED_DIR}/*IPCCommon.h
+    )
+
+    foreach(HEADER_NO_MOC_WILDCARD ${HEADERS_NO_MOC_WILDCARD})
+        file(GLOB_RECURSE GLOB_FILES ${HEADER_NO_MOC_WILDCARD})
+        list(APPEND HEADERS_NO_MOC_GLOB ${GLOB_FILES})
+    endforeach()
+
+    # but some files can be as processed as not by the moc
```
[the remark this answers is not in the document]
kunichik: done

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+    # we will analyse their text for special macro later (MOC_MACRO_NAMES)
```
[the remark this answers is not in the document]
kunichik: done

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+    set(HEADERS_CAN_MOC_WILDCARD
```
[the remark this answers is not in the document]
kunichik: moved a little lower, please, take a look.
Or do you mean drop it at all (also the foreach loop).
I added it for simplicity of adding a new line.

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+       ${CURRENT_GENERATED_DIR}/*ModulePrivate.h
+    )
+
+    set(MOC_MACRO_NAMES "Q_OBJECT" "Q_GADGET" "Q_NAMESPACE")
+
+    unset(HEADERS_CAN_MOC)
```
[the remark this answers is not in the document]
kunichik: done

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+    foreach(HEADER ${HEADERS_CAN_MOC_WILDCARD})
+        file(GLOB_RECURSE GLOB_FILES ${HEADER})
+        list(APPEND HEADERS_CAN_MOC ${GLOB_FILES})
+    endforeach()
+
+    # verify if files that can be processed by moc contains MOC_MACRO_NAMES
```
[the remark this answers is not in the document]
kunichik: done

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +428,77 @@ endmacro()
...
+    if(HEADERS_CAN_MOC)
+        foreach(HEADER_CAN_HAVE_MOC ${HEADERS_CAN_MOC})
```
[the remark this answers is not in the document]
kunichik: agree, it was a temporary decision because i've got some errors, thanks.
deleted if-statement.

cmake/faceliftMacros.cmake
kunichik: Agree!
lets HEADERS_GLOB_NO_MOC -> HEADERS_NO_MOC_GLOB etc
be for consequence:
ORDINARY | GLOB | GLOB RECURSE
-- | -- | --
HEADERS | HEADERS_GLOB | HEADERS_GLOB_RECURSE
SOURCES | SOURCES_GLOB | SOURCES_GLOB_RECURSE
HEADERS_NO_MOC | HEADERS_NO_MOC_GLOB | HEADERS_NO_MOC_GLOB_RECURSE

Facelift without extra moc processing | Current Github version
-- | --
1:27 | 1:48
1:03 | 1:05
1:20 | 1:47
1:30 | 1:08
1:10 | 1:56
0:49 | 1:07
1:03 | 1:23
1:17 | 1:33
0:52 | 1:01

cmake/faceliftMacros.cmake
kunichik: we don't care about it - all the files have been added with such code:
`    facelift_add_library(${LIBRARY_NAME}
        SOURCES_GLOB_RECURSE ${MODULE_OUTPUT_PATH}/*.cpp
        HEADERS_GLOB_RECURSE ${MODULE_OUTPUT_PATH}/*.h`
so we just have to exclude no MOC files

cmake/faceliftMacros.cmake
kunichik: you are right !

examples/MyPackage/CMakeLists.txt
kunichik: done

examples/MyPackage/CMakeLists.txt
kunichik: agree, but all the files have been added with next parameter:
`HEADERS_GLOB_RECURSE ${CMAKE_CURRENT_SOURCE_DIR}/*.h`
but if I exclude MyInterfaceCppImplementation.h with 
HEADERS_NO_MOC, I get an error 

> MyPackage/models/cpp/MyInterfaceCppImplementation.h'
  is present in both HEADERS and HEADERS_NO_MOC sections at position '0'

so either we change this example and add the files explicitly, or let it be as is.

cmake/faceliftMacros.cmake
kunichik: Great!
thanks for the tip, don with list

cmake/faceliftMacros.cmake
```diff
@@ -324,6 +324,7 @@ function(facelift_add_interface TARGET_NAME)
     facelift_add_library(${LIBRARY_NAME}
         SOURCES_GLOB_RECURSE ${MODULE_OUTPUT_PATH}/*.cpp
         HEADERS_GLOB_RECURSE ${MODULE_OUTPUT_PATH}/*.h
+        ${ARGN}
```
jacky309, line 327: remove

cmake/faceliftMacros.cmake
```diff
@@ -379,13 +381,28 @@ macro(_facelift_parse_target_arguments additionalOptions additionalOneValueArgs
...
 
     set(HEADERS ${ARGUMENT_HEADERS})
-    foreach(HEADER_GLOB ${ARGUMENT_HEADERS_GLOB_RECURSE})
-        file(GLOB_RECURSE GLOB_FILES ${HEADER_GLOB})
-        list(APPEND HEADERS ${GLOB_FILES})
+
+    set(HEADERS_NO_MOC ${ARGUMENT_HEADERS_NO_MOC})
+
+    # files for installation CANNOT be in both sections: HEADERS and HEADERS_NO_MOC
+    foreach(HEADER ${HEADERS})
```
jacky309, line 388: the same restriction applies to HEADERS and HEADERS_NO_INSTALL, and we are not checking it.

cmake/faceliftMacros.cmake
```diff
@@ -402,13 +419,43 @@ endmacro()
...
+
+    set(MOC_MACRO_NAMES "Q_OBJECT" "Q_GADGET" "Q_NAMESPACE")
+
+    # check if all the autogenerated files contain specified macros (MOC_MACRO_NAMES)
+    foreach(HEADER ${HEADERS_FOR_MOC})
+        file(READ ${HEADER} FILE_STRING)
+     
+        unset(IS_FOUND)
+        foreach(MACRO_NAME ${MOC_MACRO_NAMES})
+            string(FIND "${FILE_STRING}" "${MACRO_NAME}" IS_FOUND)
```
jacky309, line 432: You are performing that grep every time the project is reconfigured.
The code generator should generate the list of MOCed files once to avoid that.

examples/ModuleImport/CMakeLists.txt
```diff
@@ -1,11 +1,13 @@
 facelift_add_interface(MainInterface
-                       INTERFACE_DEFINITION_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/interface
-                       IMPORT_FOLDERS ${CMAKE_CURRENT_SOURCE_DIR}/interface)
+                        INTERFACE_DEFINITION_FOLDER ${CMAKE_CURRENT_SOURCE_DIR}/interface
```
jacky309, line 2: undo

cmake/faceliftMacros.cmake
kunichik: All the files have been already generated and we grep them.

Isn't this the same if we will do it here instead of grep a file by Generator after a template processing ?

cmake/faceliftMacros.cmake
kunichik: Ok, thanks

examples/ModuleImport/CMakeLists.txt
kunichik: Thanks

cmake/faceliftMacros.cmake
jacky309: The code generator is only triggered at configure time if really needed. There is no reason to perform a grep if the code generator is not triggered.

jacky309:
My suggestion was, and still is:

you can let the code generator generate a new file containing the following:
set(GENERATED_HEADERS_MOC
File1.h
File2.h
....
)
set(GENERATED_HEADERS_NO_MOC
File3.h
File4.h
....
)
set(GENERATED_SOURCES
File1.cpp
File2.cpp
....
)

In order to identify which files should be MOCed, the code generator can search for "Q_OBJECT" in the files.
Then, the implementation of the "facelift_add_interface" function can use the "include" command (https://cmake.org/cmake/help/v3.0/command/include.html) to get the definitions of those variables and call the facelift_add_library with those file lists.

cmake/faceliftMacros.cmake
```diff
@@ -211,6 +211,8 @@ function(facelift_generate_code )
             list(APPEND BASE_CODEGEN_COMMAND "--all")
         endif()
 
+        list(APPEND BASE_CODEGEN_COMMAND "--no_moc_path" "${ARGUMENT_OUTPUT_PATH}")
```
jacky309, line 214: no new parameter is needed. The file should be generated in the same folder as the generated C++ files

cmake/faceliftMacros.cmake
```diff
@@ -359,6 +361,8 @@ macro(_facelift_parse_target_arguments additionalOptions additionalOneValueArgs
     set(oneValueArgs ${additionalOneValueArgs})
     set(multiValueArgs PRIVATE_DEFINITIONS PUBLIC_DEFINITIONS PROPERTIES COMPILE_OPTIONS
         HEADERS HEADERS_GLOB HEADERS_GLOB_RECURSE
+        HEADERS_NO_MOC
+        HEADERS_NO_MOC_FILE
```
jacky309, line 365: not needed

codegen/facelift/facelift-codegen.py
```diff
@@ -320,8 +320,23 @@ def hasReturnValue(self):
...
+
+def addToNoMocList(fileName, tmpPath, path):
+    with open('{}/no_moc.cmake'.format(tmpPath), 'a') as inFile:
+         inFile.write('    {}/{}\n'.format(path, fileName))
+
+def addToNoMocIfContainsMacro(fileName, tmpPath, path):
+    if not isContainsMocMacro(fileName, tmpPath):
+        addToNoMocList(fileName, tmpPath, path)
+
+def generateFile(generator, tmpPath, templatePath, context, libraryName, libraryType,  ):
```
jacky309, line 339: generateFile should detect the nature of the generated file (whether it is a source, a header which needs to be MOCed or a header which does not require MOC) and adjust the lists of files so that the cmake file can be produced. That cmake file can be read by the facelft_add_interface function just after the code generator has been invoked.

codegen/facelift/facelift-codegen.py
```diff
@@ -355,18 +370,29 @@ def run_generation(input, output, dependency, libraryName, all):
...
     for module in system.modules:
         if module.name in modulesToGenerate:
+
             ctx.update({'module': module})
             module_path = '/'.join(module.name_parts)
             log.debug('process module %s' % module.module_name)
             ctx.update({'path': module_path})
+
             generateFile(generator, 'module/{{path}}/ModulePrivate.h', 'ModulePrivate.template.h', ctx, libraryName, "")
+            addToNoMocIfContainsMacro('module/{}/ModulePrivate.h'.format(module_path), output, noMocPath)
```
jacky309, line 386: not needed

doc/page-cmakefunctions.h
```diff
@@ -46,6 +46,8 @@ facelift_add_qml_plugin(<TargetName>
                         [HEADERS_GLOB [Globbing expressions]...]
                         [SOURCES_GLOB_RECURSE [Globbing expressions]...]
                         [HEADERS_GLOB_RECURSE [Globbing expressions]...]
+                        [HEADERS_NO_MOC hdr1 [hdr2...]]
+                        [HEADERS_NO_MOC_FILE file]
```
jacky309, line 50: not needed

doc/page-cmakefunctions.h
```diff
@@ -76,7 +80,25 @@ facelift_add_library(<TargetName>
 \endcode
 
 This function adds a library with the name \e TargetName from the specified source and header files. The parameter descriptions can be found below.
-
+In the parameter HEADERS_NO_MOC_FILE you can set the file with a list of the files which sould be excluded from processing by moc.
+This set should have the name HEADERS_NO_MOC_FROM_FILE. E.g in we create the file ~/nomoc.cmake and fill it like this:
```
jacky309, line 84: HEADERS_NO_MOC is the only parameter which needs to be documented

cmake/faceliftMacros.cmake
bitmouse, line 365: HEADERS_NO_MOC_FILE is one value arg, should go there

> not needed

It is needed in our case AFAIK. To be confirmed yet.

codegen/facelift/facelift-codegen.py
```diff
@@ -320,8 +320,23 @@ def hasReturnValue(self):
...
-
-def generateFile(generator, outputPath, templatePath, context, libraryName, libraryType):
+def isContainsMocMacro(fileName, tmpPath):
+    mocMacro = ["Q_OBJECT", "Q_GADGET", "Q_NAMESPACE"]
+    with open('{0}/{1}'.format(tmpPath, fileName), 'r') as inFile:
+        for line in inFile:
+            if any(word in line for word in mocMacro):
+                return True
+    return False
+
```
bitmouse, line 330: grep again?

codegen/facelift/facelift-codegen.py
bitmouse, line 386: at this point you can introspect local data and tell if there is MOC or no MOC. It will be faster than grepping file.

bitmouse (changes requested):
Remove grep (in any form) please!

codegen/facelift/facelift-codegen.py
jacky309: Not sure what "introspect" means here, and what "local data" means.
The "generateFile" function can use some kind of grep on the template file, in order to detect if the generated file needs to be MOCed or not. The result of that grep can be cached so that no further grep will be needed when generating a file based on the same template.
Not sure whether this is the same idea as @bitmouse had.

cmake/faceliftMacros.cmake
kunichik: needed for lists created by applications that can use facelift's facelift_add_library

codegen/facelift/facelift-codegen.py
kunichik: deleted in the new version...

doc/page-cmakefunctions.h
kunichik: needed

doc/page-cmakefunctions.h
kunichik: HEADERS_NO_MOC_FROM_FILE needed, please see  comment above.

codegen/facelift/facelift-codegen.py
kunichik: fixed in the new version

cmake/faceliftMacros.cmake
kunichik: this is needed for correct paths in HEADERS_NO_MOC_FROM_FILE

codegen/facelift/facelift-codegen.py
kunichik: @bitmouse has meant and yes - as I've just figured out - we can inspect the model's "size" and add the file to the NO_MOC list if there are no any structures in the model.

codegen/facelift/facelift-codegen.py
kunichik: fixed

codegen/facelift/facelift-codegen.py
kunichik: approach has been changed in the new version

codegen/facelift/facelift-codegen.py
bitmouse: Grep is slow, we are going to avoid it whenever possible. If we can use something that works in constant time O(1), or even O(log n) then we will prefer that over grep, which is O(n). EOT.

doc/page-cmakefunctions.h
```diff
@@ -114,6 +136,8 @@ This function adds a test with the name \e TargetName.The parameter descriptions
 \param SOURCES_GLOB_RECURSE List of source files searched recursively inside folder and subfolders, used to build this target provided as a wildcard expression
 \param HEADERS_GLOB_RECURSE List of header files to be MOCed and added to this target, searched recursively inside folder and subfolders provided as a wildcard expression
 \param LINK_LIBRARIES List of libraries that will be linked with this target
+\param HEADERS_NO_MOC The headers that should be excluded from moc processing
+\param HEADERS_NO_MOC_FILE The list of the files that should be excluded from moc processing
```
jacky309, line 140: If you introduce that , you also need to introduce HEADERS_FILE, HEADERS_NO_INSTALL_FILE.
But what problem is it supposed to solve ? Did you experience any limitation with CMake regarding lists length ??

codegen/facelift/facelift-codegen.py
jacky309: Here is how "slow" grep is:
`
$ time grep 'Q_OBJECT\|Q_ENUM\|Q_GADGET' ~/facelift/codegen/facelift/templates/*
/home/jacques/facelift/codegen/facelift/templates/Enum.template.h:    Q_GADGET
/home/jacques/facelift/codegen/facelift/templates/Enum.template.h:    Q_ENUM(Type)
/home/jacques/facelift/codegen/facelift/templates/ImplementationBaseQML.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ImplementationBaseQML.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ImplementationBase.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCAdapter.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCDBusServiceAdapter.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCProxyAdapter.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCProxy.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCProxy.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/IPCServiceAdapter.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ModulePrivate.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/QMLAdapter.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ServiceMonitor.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ServiceNotAvailableImpl.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/Service.template.h:    Q_GADGET
/home/jacques/facelift/codegen/facelift/templates/Service.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/ServiceWrapper.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/StructQObjectWrapper.template.h:    Q_OBJECT
/home/jacques/facelift/codegen/facelift/templates/Struct.template.h:    Q_GADGET

real	0m0,002s
user	0m0,001s
sys	0m0,001s
`

bitmouse: Sorry, but this answer must be a joke. If you are trying to say that O(n) scales as good as O(1) this means that you completely do not understand it.
Now multiply that x1000. And 100k. And 1M. You can easily calculate this calculate that. More precisely grep will do O(k*n) where k is number of files and n number of bytes in the file. Checking the structure/list size on the generator side works in O(k) time, because we do it only once per file. 
I can only recommend 'Algorithms' by Robert Sedgewick and Kevin Wayne or another book 'Introduction to algorithms' by Thomas Cormen.

doc/page-cmakefunctions.h
kunichik: >HEADERS_NO_INSTALL_FILE, SOURCES_FILE, etc....

sure, if we need it, then why not, but HEADERS_NO_MOC_FILE is enough for now.

> But what problem is it supposed to solve ? Did you experience any limitation with CMake regarding lists length ??

in this PR i'm processing and iterating HEADERS_NO_MOC_GENERATED
and removing such files from HEADERS_FOR_MOC.
But what if the number of non-MOC files is too large and we use **facelift_add_library** outside of Facelift, where can we define this list?
File is the best choice, and no extra options like HEADERS_NO_INSTALL_FILE are needed !

codegen/facelift/facelift-codegen.py
```diff
@@ -320,8 +320,11 @@ def hasReturnValue(self):
 setattr(qface.idl.domain.Property, 'cppMethodArgumentType', property(cppMethodArgumentType))
 
 ##############################
+def addToNoMocList(fileName, tmpPath, path):
+    with open('{}/no_moc.cmake'.format(tmpPath), 'a') as inFile:
+         inFile.write('    {}/{}\n'.format(path, fileName))
 
-def generateFile(generator, outputPath, templatePath, context, libraryName, libraryType):
+def generateFile(generator, tmpPath, templatePath, context, libraryName, libraryType,  ):
```
bitmouse, line 327: remove , at the end

codegen/facelift/facelift-codegen.py
```diff
@@ -363,10 +369,17 @@ def run_generation(input, output, dependency, libraryName, all):
             log.debug('process module %s' % module.module_name)
             ctx.update({'path': module_path})
             generateFile(generator, 'module/{{path}}/ModulePrivate.h', 'ModulePrivate.template.h', ctx, libraryName, "")
+            if not module.structs:
+                addToNoMocList('module/{}/ModulePrivate.h'.format(module_path), output, noMocPath)
+
```
bitmouse, line 374: In fact you open noMocPath multiple times. You could optimize it out to open once and pass a handle instead of path. This should be a bit faster (I don't know if this optimization is worth something).

codegen/facelift/facelift-codegen.py
kunichik: done

codegen/facelift/facelift-codegen.py
kunichik: agree.
it's not noticeable now, but if there will be e.g. 4000 line insertion it is very slow (I checket it out).
So yep - changed. The file opens only once and closes after leaving the scope.

codegen/facelift/facelift-codegen.py
```diff
@@ -320,8 +320,10 @@ def hasReturnValue(self):
 setattr(qface.idl.domain.Property, 'cppMethodArgumentType', property(cppMethodArgumentType))
 
 ##############################
+def addFileNameToNoMocList(noMocListFile, filePath, fileName):
+    noMocListFile.write('    {}/{}\n'.format(filePath, fileName))
```
bitmouse, line 324: https://docs.python.org/3/library/os.path.html#os.path.join

kunichik: done

bitmouse (approved):

codegen/facelift/facelift-codegen.py
jacky309: What about reading my comments, before providing such irrelevant answers ??

bitmouse: In which part my answer is irrelevant?

PR #309, opened 2020-10-13
Drop Windows OS support

kunichik:


cmake/faceliftMacros.cmake
```diff
@@ -975,15 +958,13 @@ function(facelift_add_qml_plugin PLUGIN_NAME)
...
-        # not supported for now on Win32 since the required libraries can't be loaded without setting the PATH variable
-        add_custom_command(
-            OUTPUT  ${CMAKE_BINARY_DIR}/${INSTALL_PATH}/plugins.qmltypes
-            COMMAND ${_qt5Core_install_prefix}/bin/qmlplugindump -noinstantiate ${URI} ${PLUGIN_MAJOR_VERSION}.${PLUGIN_MINOR_VERSION} ${CMAKE_BINARY_DIR}/imports -output ${CMAKE_BINARY_DIR}/${INSTALL_PATH}/plugins.qmltypes || touch ${CMAKE_BINARY_DIR}/${INSTALL_PATH}/plugins.qmltypes
-            DEPENDS ${PLUGIN_NAME}
-        )
-        add_custom_target("generate_qmltypes_${PLUGIN_NAME}" ALL DEPENDS ${CMAKE_BINARY_DIR}/${INSTALL_PATH}/plugins.qmltypes)
-        install(FILES ${CMAKE_BINARY_DIR}/${INSTALL_PATH}/plugins.qmltypes DESTINATION ${INSTALL_PATH})
-    endif()
+    add_custom_command(
```
bitmouse, line 961: Is it safe to remove NOT CMAKE_CROSSMOPLILING?

cmake/faceliftMacros.cmake
```diff
@@ -203,10 +197,6 @@ function(facelift_generate_code )
             list(APPEND BASE_CODEGEN_COMMAND "--dependency" "${IMPORT_FOLDER}")
         endforeach()
 
-        if(ARGUMENT_LIBRARY_NAME)
-            list(APPEND BASE_CODEGEN_COMMAND "--library" "${ARGUMENT_LIBRARY_NAME}")
-        endif()
-
```
bitmouse, line 209: what does it change?

cmake/faceliftMacros.cmake
```diff
@@ -232,10 +222,6 @@ function(facelift_generate_code )
             ERROR_VARIABLE CODEGEN_ERROR
         )
 
-        if(LANG_SET_BY_FACELIFT)
-            unset(ENV{LANG})
-        endif()
-
```
bitmouse, line 238: what does it change?

bitmouse (changes requested):

cmake/faceliftMacros.cmake
kunichik: see former codegen parameter library

cmake/faceliftMacros.cmake
kunichik: it was introduced for WIN32
> not supported for now on Win32

wasn't it ?

cmake/faceliftMacros.cmake
kunichik: undo

cmake/faceliftMacros.cmake
bitmouse: I did not check TBH. However there is a logical AND, therefore if we drop the windows support, we have it always true, the only part left for checking is CMAKE_CROSSCOMPILING. We can still cross compile it for e.g. Arm or other embedded. If you prove this changes nothing, I approve PR.

kunichik: agree.
I could delete only "NOT WIN32" here, let CMAKE_CROSSCOMPILING be

bitmouse (approved):

PR #313, opened 2020-10-14
Add auxiliary API for Proxy

zusage:


src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -54,7 +54,15 @@ class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
...
-
+    // Set observers
+    void setObservers(std::vector<IObserver *> &observers){
+        for(auto observer: observers){
+            auto connection = std::make_shared<QMetaObject::Connection>();
+            *connection = QObject::connect(this, &InterfaceBase::readyChanged, observer, [observer, connection](){
+                observer->onReadyChanged( connection );
+            });
+        }
+    }
```
bitmouse: This should be inside IsReadyObserver class

bitmouse (changes requested):
Please add unit tests or/and test showing the usage of new functionality.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+#include <memory>
+#include <QObject>
+
+namespace facelift {
+
+class IObserver : public QObject
+{
+    Q_OBJECT
+public:
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection ) = 0;
```
kunichik: delete the space between connection and the bracket 
'connection ) = 0;
'

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+};
+
+// Single-time observer which will unregister itself when done
+template<typename Function>
+class SingleTimeObserver : public IObserver
+{
+    Function m_func;
+public:
+    explicit SingleTimeObserver(Function func): m_func{func} {}
```
kunichik: add a space before colon

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+    ~SingleTimeObserver()=default;
```
kunichik: add spaces before and after equal sign

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) override {
+         m_func();
+         QObject::disconnect(*connection);
+    }
+};
```
kunichik: add a new line

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+// Standard observer which will work for each signal
+template<typename Function>
+class StandartObserver : public IObserver
+{
+    Function m_func;
+public:
+    explicit StandartObserver(Function func): m_func{func} {}
```
kunichik: as above

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+    ~StandartObserver()=default;
```
kunichik: as above

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection ) override {
+        Q_UNUSED(connection)
+        m_func();
+    }
+};
+}
```
kunichik, line 196: add a new line

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+    Function m_func;
+public:
+    explicit StandartObserver(Function func): m_func{func} {}
+    ~StandartObserver()=default;
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection ) override {
+        Q_UNUSED(connection)
+        m_func();
+    }
+};
```
kunichik, line 194: add a new line

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+template<typename Function>
+class StandartObserver : public IObserver
+{
+    Function m_func;
+public:
+    explicit StandartObserver(Function func): m_func{func} {}
+    ~StandartObserver()=default;
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection ) override {
+        Q_UNUSED(connection)
```
kunichik: delete Q_UNUSED macro

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,72 @@
...
+// Standard observer which will work for each signal
+template<typename Function>
+class StandartObserver : public IObserver
+{
+    Function m_func;
+public:
+    explicit StandartObserver(Function func): m_func{func} {}
+    ~StandartObserver()=default;
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection ) override {
```
kunichik: delete parameter name

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -54,7 +54,15 @@ class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
     IPCProxyBase(QObject *parent) : AdapterType(parent)
     {
     }
-
+    // Set observers
+    void setObservers(std::vector<IObserver *> &observers){
+        for(auto observer: observers){
```
kunichik: add a space before the colon and a space before the curly braces

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -54,7 +54,15 @@ class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
     IPCProxyBase(QObject *parent) : AdapterType(parent)
     {
     }
-
```
kunichik: don't delete this line

kunichik (changes requested):

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+    int m_value;
+};
+
+class IPCProxyBaseTest : public testing::Test
+{
+public:
+	Counter c1;
+	Counter c2;
+	StandartObserver  < std::function<void()> > * obs1 = new StandartObserver  < std::function<void()> >(std::bind(&Counter::setValue1, &c1) );
+	SingleTimeObserver< std::function<void()> > * obs2 = new SingleTimeObserver< std::function<void()> >(std::bind(&Counter::setValue2, &c2) );
```
bitmouse: mem leak x 2

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+public:
+    Counter() { m_value = 0; }
+	int value() const { return m_value; }
+
+    void setValue1(){
+	    qDebug() << "Counter setValue1() ";
+	}
+	void setValue2(){
+	    qDebug() << "Counter setValue2() ";
```
bitmouse: not needed

bitmouse (changes requested):
I do not understand the test. Are you testing setter/getter pair?

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,52 @@
...
+TEST_F(IPCProxyBaseTest, setObservers)
+{
+    const auto expected = std::vector<IObserver*>{obs1, obs2,};
+    proxyBase.m_readyObserver.setObservers(expected);
+    const auto actual = proxyBase.m_readyObserver.getObservers();
+    EXPECT_EQ(expected.size(), actual.size());
+    EXPECT_EQ(expected[0], actual[0]);
+    EXPECT_EQ(expected[1], actual[1]);
+}
+
```
bitmouse: What do you test?
Could you test the actual usage of new functionality?

zusage: This is not a complete test cases. I will inform you. 

Wait, please.

src/ipc/ipc-common/observer.h
kunichik: Please add new lines as asked above.
It is nessecary at the end of a file at least.
You could set your IDE to paste a new line at the end if you have forgotten to do it.

kunichik (changes requested):
Please align the code ( curly braces etc)

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) = 0;
+};
+
+class IsReadyObserver: public QObject
+{
+    Q_OBJECT
+
+    std::vector<IObserver *> m_observers;
+    std::shared_ptr<QMetaObject::Connection> connection;
+public:
```
kunichik, line 87: add the new linebefore public:

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) = 0;
+};
+
+class IsReadyObserver: public QObject
+{
+    Q_OBJECT
+
+    std::vector<IObserver *> m_observers;
+    std::shared_ptr<QMetaObject::Connection> connection;
```
kunichik: rename connection -> m_connection

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+#include <vector>
+#include <QObject>
+
+namespace facelift {
+
+class IObserver : public QObject
+{
+    Q_OBJECT
+public:
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) = 0;
```
kunichik: why do you use std::shared_ptr ?
Isn't it enough to use std::unique_ptr + std::make_unique ?
Of course, this also applies to other parts of PR.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+{
+    Q_OBJECT
+
+    std::vector<IObserver *> m_observers;
+    std::shared_ptr<QMetaObject::Connection> connection;
+public:
+    IsReadyObserver() : connection{std::make_shared<QMetaObject::Connection>()}
+    {
+        *connection = QObject::connect(this, &IsReadyObserver::readyChanged, this, [this] () {
+            for (auto observer : m_observers) {
```
kunichik: how about UB here ?
consider such example (sorry for no format - there is something wrong with github parser):

struct Base
{
    virtual void foo() = 0;
};

class Example : public Base {
public:
    void foo() override { std::cout << "foo ! value = " << m_value << '\n';}
      
private:
    int m_value {};
};


int main()
{
    std::vector<Base*> m_examples;
    
    m_examples.push_back(new Example());
    m_examples.push_back(nullptr);  
    m_examples.push_back(new Example());
    
    for(auto example : m_examples) {
        std::cout << "running foo \n";
        //if(example /*!= nullptr*/)
            example->foo();
    }

    return 0;
}

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+        });
+    }
+
+    // Set observers
+    void setObservers(const std::vector<IObserver *> &observers) {
+        m_observers = observers;
+    }
+
+    // Get observers
+    const std::vector<IObserver *> &getObservers() const {
```
kunichik: Have you considered using QVector instead from the point of view the place where we would create the vector ?

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+    void onReadyChanged() {
+        emit readyChanged();
+    }
+};
+
+// Single-time observer which will unregister itself when done
+template<typename Function>
+class SingleTimeObserver : public IObserver
+{
+    Function m_func;
```
kunichik: do not use short names.
m_func-> m_function.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,108 @@
...
+
+    Q_SIGNAL void readyChanged();
+
+    void onReadyChanged() {
+        emit readyChanged();
+    }
+};
+
+// Single-time observer which will unregister itself when done
+template<typename Function>
```
kunichik: use T for the type instead

tests/unittest_observer/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+
+find_package(GTest)
+if(${GTEST_FOUND})
+    find_library(GMOCK_LIBRARY gmock REQUIRED HINTS "${CMAKE_SYSTEM_PREFIX_PATH}")
+    if(NOT ${GMOCK_LIBRARY} STREQUAL "GMOCK_LIBRARY-NOTFOUND")
```
kunichik: verify if it's enough to use if(GMOCK_LIBRARY-NOTFOUND) instead and reverse logic

tests/unittest_observer/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+    find_package(Threads REQUIRED)
+    set(FACELIFT_GTEST_LIBRARIES ${GTEST_BOTH_LIBRARIES} GTest::GMock Threads::Threads)
+    include_directories(${GTEST_INCLUDE_DIRS})
+
+    facelift_add_test(UnitTestsObserver
+        SOURCES FaceliftObserverTest.cpp
+        LINK_LIBRARIES ${FACELIFT_GTEST_LIBRARIES} FaceliftIPCCommonLib FaceliftModelLib)
+
+else()
+    message(WARNING "Required package google test not found!")
```
kunichik: google test package is required but the message is only Warning, is it enough ?

tests/unittest_observer/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
...
+        message(WARNING "Google test/mock not found.")
+    endif()
+
+    find_package(Threads REQUIRED)
+    set(FACELIFT_GTEST_LIBRARIES ${GTEST_BOTH_LIBRARIES} GTest::GMock Threads::Threads)
+    include_directories(${GTEST_INCLUDE_DIRS})
+
+    facelift_add_test(UnitTestsObserver
+        SOURCES FaceliftObserverTest.cpp
+        LINK_LIBRARIES ${FACELIFT_GTEST_LIBRARIES} FaceliftIPCCommonLib FaceliftModelLib)
```
kunichik: use a new line for each value of LINK_LIBRARIES

kunichik (changes requested):

src/ipc/ipc-common/observer.h
kunichik: without nullptr check we have UB

src/ipc/ipc-common/IPCProxyBase.h
```diff
@@ -49,10 +49,12 @@ class IPCProxyBase : public AdapterType, protected IPCProxyBaseBase
 
 public:
     using InterfaceType = AdapterType;
+    IsReadyObserver m_readyObserver{};
```
kunichik: Is this necessary to be public ?

src/ipc/ipc-common/observer.h
kunichik: change lines 50 and 51!

src/ipc/ipc-common/observer.h
kunichik: so why?

src/ipc/ipc-common/observer.h
kunichik: done, please write about this because it's pretty hard to navigate in squashed commits.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,107 @@
...
+    Q_OBJECT
+
+    QVector<IObserver *> m_observers{};
+public:
+    IsReadyObserver() {}
+
+    // Set observers
+    void setObservers(const QVector<IObserver *> &observers) {
+        m_observers = observers;
+        for(auto observer: observers){
```
kunichik: UB

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,107 @@
...
+            auto connection = std::make_shared<QMetaObject::Connection>();
+            *connection = QObject::connect(this, &IsReadyObserver::readyChanged, observer, [observer, connection](){
+                observer->onReadyChanged( connection );
+            });
+         }
+    }
+
+    // Get observers
+    const QVector<IObserver *> &getObservers() const {
```
kunichik: Why do we use QVector here, is it better by some reason ?
Please, answer previous comments, because it's quite hard to compare and remember than.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,107 @@
...
+
+    Q_SIGNAL void readyChanged();
+
+    void onReadyChanged() {
+        emit readyChanged();
+    }
+};
+
+// Single-time observer which will unregister itself when done
+template<typename F>
```
kunichik: Use T instead - it's common type convention.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,107 @@
...
+class SingleTimeObserver : public IObserver
+{
+    F m_function;
+public:
+    explicit SingleTimeObserver(F func) : m_function{func} {}
```
kunichik: func->function

tests/unittest_observer/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+    find_package(Threads REQUIRED)
+    if(${THREADS_NOTFOUND})
+        message(ERROR "Required package Threads not found.")
+    endif()
+    set(FACELIFT_GTEST_LIBRARIES ${GTEST_BOTH_LIBRARIES} GTest::GMock Qt5::Test Threads::Threads)
+    include_directories(${GTEST_INCLUDE_DIRS})
+
+    facelift_add_test(UnitTestsObserver
+        SOURCES FaceliftObserverTest.cpp
+        LINK_LIBRARIES ${FACELIFT_GTEST_LIBRARIES} FaceliftIPCCommonLib)
```
kunichik: new line for each value

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+private:
+    int m_value;
+};
+
+class IPCProxyBaseTest : public ::testing::Test
+{
+public:
+    Counter c1;
+    Counter c2;
+    StandartObserver  < std::function<void()> > * obs1 = new StandartObserver  < std::function<void()> > (std::bind(&Counter::incValue, &c1) );
```
kunichik: use auto

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+using namespace facelift;
+
+class Counter
+{
+public:
+    Counter() { m_value = 0; }
+
+    void incValue() {
+        m_value++;
```
kunichik: why postfix increment ?

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+public:
+    Counter c1;
+    Counter c2;
+    StandartObserver  < std::function<void()> > * obs1 = new StandartObserver  < std::function<void()> > (std::bind(&Counter::incValue, &c1) );
+    SingleTimeObserver< std::function<void()> > * obs2 = new SingleTimeObserver< std::function<void()> > (std::bind(&Counter::incValue, &c2) );
+
+    IPCProxyBase<InterfaceBase> proxyBase{nullptr};
+
+    ~IPCProxyBaseTest() {
+        delete obs1;
```
kunichik: Do you have some specific reason for manual memory management ?

kunichik (commented):
I ask you reply all the comments!
At least: "Done", "not needed", "to be done", or answer to the question.
Otherwise I cannot see any reason for wasting the time and doing reviews when 
your comments picked selectively and yours questions are ignored!

src/ipc/ipc-common/IPCProxyBase.h
zusage: > I ask you reply all the comments!
> At least: "Done", "not needed", "to be done", or answer to the question.
> Otherwise I cannot see any reason for wasting the time and doing reviews when
> your comments picked selectively and yours questions are ignored!

Sorry me, but for save time, I try to make corrections without comment.

tests/unittest_observer/CMakeLists.txt
zusage: Changed to "ERROR"

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: "add the new linebefore public:" 
Done.

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: This is just for test example.

src/ipc/ipc-common/observer.h
zusage: Using of QVector as recommedation was noticed at previous comment:
"Have you considered using QVector instead from the point of view the place where we would create the vector ?"
I am currently using QVector for uniform use main types.

src/ipc/ipc-common/IPCProxyBase.h
kunichik: no problem, otherwise it takes my time instead for finding older comments and compare with new version. Obvious cases you could mark as resolved as well.
I mean:
> change the variable name func -> functor
> Done

and mark as resolved or just only mark as resolved but the Done  comment will help.
I think we'll do review without squash in the future.

src/ipc/ipc-common/observer.h
kunichik: it was a question:
> Have you considered using QVector... ?

But why is it should be used here ?
E.g. Implicit Sharing matters and so on...

>I am currently using QVector for uniform use main types.

so does QVector used from the call context ?

src/ipc/ipc-common/observer.h
kunichik: thanks, you could delete the empty line between Q_OBJECT and QVector and see how's better.
that is all

src/ipc/ipc-common/IPCProxyBase.h
bitmouse: Why is it here? We discussed that already that we do not want to extend this class.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,109 @@
...
+#include <QVector>
+#include <QObject>
+
+namespace facelift {
+
+class IObserver : public QObject
+{
+    Q_OBJECT
+public:
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) = 0;
```
bitmouse: IObserver::onReadyChanged? I think this method should be in IsReadyObserver.

tests/unittest_observer/FaceliftObserverTest.cpp
kunichik: Do you mean that it is not final edition or it is doesn't matter what to write in test examples ?

tests/unittest_observer/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
+
```
bitmouse: Please move tests to tests/unittest which is already present. You do not want to reinvent the wheel, nether copy the code.

tests/unittest_observer/FaceliftObserverTest.cpp
bitmouse: std::unique_ptr? 
Why not just make it a member of the class (not a pointer)?

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    // Set observers to proxy
+    const auto expected = QVector<IObserver*>{obs1, obs2,};
+    proxyBase.m_readyObserver.setObservers(expected);
+    const auto actual = proxyBase.m_readyObserver.getObservers();
+
+    // Check installed observers
+    EXPECT_EQ(expected.size(), actual.size());
+    EXPECT_EQ(expected[0], actual[0]);
+    EXPECT_EQ(expected[1], actual[1]);
```
bitmouse: I don't get this part. Are you testing setter/getter pair?

tests/unittest_observer/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+    ASSERT_EQ(spy.count(), 0);
+    proxyBase.readyChanged();
+    proxyBase.readyChanged();
+    proxyBase.readyChanged();
+    ASSERT_EQ(spy.count(), 3);
+
+    // Check values after signal call
+    ASSERT_EQ(c1.getValue(), 3); // for StandartObserver
+    ASSERT_EQ(c2.getValue(), 1); // for SingleTimeObserver
+}
```
bitmouse: You do not test the functionality. Also I am not quite sure how to use your observer, which was meant to be similar to discussed "callOnceReady" and "callOnReady".

bitmouse (changes requested):

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: In this case it is no matter, result will be a same. But I agree the postfix is more expensive, therefore necessary to change to **prefix** increment.

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: I suppose that writing in test cases should be as simple as possible with obviously stages.

tests/unittest_observer/CMakeLists.txt
pfagner: Okay.

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: This is just an additional check. I will remove these checks.

tests/unittest_observer/FaceliftObserverTest.cpp
kunichik: I'm not sure (and do not argue) that it doesn't matter, but anyway - it's better with the prefix form

tests/unittest_observer/FaceliftObserverTest.cpp
kunichik: not always!
I agree with some examples like  that 

int addition(int a, int b);
int x1 = addition(1,2);
int x2 = addition(2,3);
EXPECT_EQ(3, x1);
EXPECT_EQ(5, x2);

where it is in one small place, but when it is inside some class, and may result to memory leaks, exposed in the public repository, I don't agree that test can be justification
Anyway it looks messy. 
And what if someone decides to add
another pointer and forget to delete it, or
will add an array of pointers?

And btw
using smart pointers, auto keyword, range-based loops etc is easier
both - for reading and writing.

src/ipc/ipc-common/observer.h
kunichik: No UB here but the third connect parameter looks after that and warn about that like this
**QObject::connect: invalid null parameter**
see https://medium.com/genymobile/how-c-lambda-expressions-can-improve-your-qt-code-8cd524f4ed9f

I think we'll get rid of that warning anyway and verify if an observer variable is nullptr.

tests/unittest_observer/CMakeLists.txt
zusage: Done

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: Done

tests/unittest_observer/CMakeLists.txt
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: On October 16, 2020 we talked with you about possible use QVector instead std::vector.
Before this moment I have not considered using QVector and have not consulted anyone.
In the last commit, I used QVector instead of std::vector to evaluate the work, performance and for estimation using in future.
I suppose that using QVector is more better way than std::vector because it is full compatible with Qt, but possible it have a constraint or weak sides.
Q: "so does QVector used from the call context ?"
A: Yes.

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: Unfortunately auto is not supported.
In this case, we have some limitation knowing as AAA Style (Almost Always Auto):
https://www.bfilipek.com/2015/02/non-static-data-members-initialization.html

src/ipc/ipc-common/observer.h
zusage: To avoid that, I pass a “context” argument to the connect() call.
It won't affect the execution of our lambda, but when 'observer' is deleted, Qt will notice and will disconnect from lambda.
for(auto observer: observers){
    auto connection = std::make_shared<QMetaObject::Connection>();
    *connection = QObject::connect(this, &IsReadyObserver::readyChanged, **observer**, [observer, connection](){
         observer->onReadyChanged( connection );
      });

tests/unittest_observer/FaceliftObserverTest.cpp
kunichik: agree
**ISO C++ forbids in-class initialization of non-const static member** 
it was misleading, thought i in some function.

kunichik: please make this line less long, 
e.g. this thing "> >" is not needed anymore

StandartObserver  < 
one space is spare

delete extra spaces in all this line like this:
`>>(std::bind(&Counter::incValue, &c1)); `

you could use using keyword as well
and as mentioned earlier use unique_ptr or class member.

src/ipc/ipc-common/observer.h
zusage: Agree. To be done.

kunichik: what happens if the **observer** has already been deleted or is nullptr **at the moment** of connect ?
we'll get runtime warning
**QObject::connect: invalid null parameter**
lets get rid of this !

tests/unittest/CMakeLists.txt
```diff
@@ -6,18 +6,34 @@ if(${GTEST_FOUND})
...
+    endif()
+    set(FACELIFT_GTEST_LIBRARIES ${GTEST_BOTH_LIBRARIES} GTest::GMock Qt5::Test Threads::Threads)
     include_directories(${GTEST_INCLUDE_DIRS})
 
     facelift_add_test(UnitTests
         SOURCES FaceliftUtilsTest.cpp
-        LINK_LIBRARIES ${FACELIFT_GTEST_LIBRARIES} FaceliftCommonLib)
+        LINK_LIBRARIES
+	${FACELIFT_GTEST_LIBRARIES}
+	FaceliftCommonLib
```
kunichik: alignment

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,110 @@
...
+        m_observers = observers;
+        for(auto observer: observers){
+            if(observer){
+                auto connection = std::make_shared<QMetaObject::Connection>();
+                *connection = QObject::connect(this, &IsReadyObserver::readyChanged, observer, [observer, connection](){
+                    observer->onReadyChanged( connection );
+                });
+            }
+         }
+    }
```
kunichik, line 124: add spaces after brackets, for, if etc.
see alignment of curly braces (lines 64, 65)

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,110 @@
...
+template<typename T>
+class StandartObserver : public IObserver
+{
+    T m_function;
+
+public:
+    explicit StandartObserver(T function) : m_function{function} {}
+    ~StandartObserver() = default;
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection>  ) override {
```
kunichik: extra spaces after Connection>

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,111 @@
...
+{
+    Q_OBJECT
+public:
+    virtual void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) = 0;
+};
+
+class IsReadyObserver: public QObject
+{
+    Q_OBJECT
+    QVector<IObserver *> m_observers{};
```
kunichik: do not use raw pointers for observers.
consider QPointer, shared_ptr or weak_ptr

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,111 @@
...
+template<typename T>
+class SingleTimeObserver : public IObserver
+{
+    T m_function;
+
+public:
+    explicit SingleTimeObserver(T function) : m_function{function} {}
+    ~SingleTimeObserver() = default;
+
+    void onReadyChanged(std::shared_ptr<QMetaObject::Connection> connection) override {
```
kunichik: use const ref/ref for heavyweight objects.
Do not pass them by value - it is costly enough

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,111 @@
...
+    QVector<IObserver *> m_observers{};
+
+public:
+    IsReadyObserver() {}
+
+    // Set observers
+    void setObservers(const QVector<IObserver *> &observers) {
+        m_observers.clear();
+        for(auto observer: observers) {
+            if(observer) {
```
kunichik: I think that nullptr inside the observers would be error.
Use Q_ASSERT instead.

zusage: The expression inside the Q_ASSERT will **not** be evaluated in non-debug build configurations.
https://doc.qt.io/qt-5/qtglobal.html#Q_ASSERT

src/ipc/ipc-common/observer.h
zusage: Done

tests/unittest/CMakeLists.txt
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

src/ipc/ipc-common/observer.h
kunichik: yes, but the main goal to find the segfault place e.g.
it's better Q_ASSTERT in debug than skip this and get an error in other place.

src/ipc/ipc-common/IPCProxyBase.h
bitmouse: Please move it from this class. We do not want it here!

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: Done.

src/ipc/ipc-common/IPCProxyBase.h
zusage: Done

src/ipc/ipc-common/observer.h
zusage: Done

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: I added comments for the test, which briefly describes the sequence of work and interaction between "Proxy", "IsReadyObserver" and observers for "callOnReady", "callOnceReady". Instead "Proxy" there may be a different type and signal.

kunichik (commented):
add observer.cpp and move an appropriate code there

tests/unittest_observer/FaceliftObserverTest.cpp
zusage: The main goal is to have a flexible service, present as independent/extern component, that reacts on certain states of one object of inspection and execute vary callbacks either once or permanent.
In general it is intend for a Qt objects a certain kind of types, that has the "ready" state, it is necessary to provide the ability to track this state using the "IsReadyObserver" component and an arbitrary number of observers of two types: "callOnReady" and "callOnceReady", which are callbacks. The ready state is determined by a specific signal that emitted from inspected object and "ready" state. The "IsReadyObserver" is the watcher of the signal of interest in "Proxy" and contain observers with callback to execute.
**INSTALLATION**: the user creates an "IsReadyObserver" and sets in it: the object's address, the object's interface signal and the object's readiness function. After that set/load observers to "IsReadyObserver".
For example:
`readyObserver.watch( &proxy, &IPCProxy<InterfaceBase>::readyChanged, &IPCProxy<InterfaceBase>::ready );`
**CONDITION FOR PROCESSING**: for the inspected object, the interface signal has changed, while the ready property is in the "ready" state. In this case, the callback in observers will be executed once.
If the interface signal for the inspected object has changed, but the ready property is in the "not ready" state. In this case, the callback in observers will not executed.
**REQUIREMENT**: One change in the interface signal and "ready" state corresponds to a single execution of the callback. No signal can be missed. 

Main steps:
- Creation of main objects: object for inspection "Proxy" (this is one of more possible types), object for observation "IsReadyObserver", objects for executing "Observers".
- Installing objects/function to be executed in observers depending on the need: "callOnReady" or "callOnceReady".
- Setting up to "IsReadyObserver" to watch out for signal in "Proxy".
- Catching signal and execution of callback.
- Completion of work.

Testing:
1) Functionality check and main life cycle of components (creation, execution, completion)
2) Processing of erroneous or incorrect data
3) Memory leaks and undefined behavior
4) Load testing for performance (performance, latency, metrics).
5) Stress testing for failure

Possible points 1-4 are minimal necessary.

src/ipc/ipc-common/observer.h
```diff
@@ -0,0 +1,129 @@
...
+#include <QVector>
+#include <QObject>
+
+namespace facelift {
+
+class IObserver : public QObject
+{
+    Q_OBJECT
+public:
+    virtual void IsReadyObserver(const std::shared_ptr<QMetaObject::Connection> &connection) = 0;
```
bitmouse: What is this method? Why?

tests/unittest/CMakeLists.txt
```diff
@@ -6,18 +6,34 @@ if(${GTEST_FOUND})
         add_library(GTest::GMock UNKNOWN IMPORTED)
         set_target_properties(GTest::GMock PROPERTIES IMPORTED_LOCATION ${GMOCK_LIBRARY})
     else()
-        message(WARNING "Google test/mock not found.")
+        message(ERROR "Google test/mock not found.")
```
bitmouse: It should be warning! If there are not gtest installed, we just ignore these!

tests/unittest/CMakeLists.txt
```diff
@@ -6,18 +6,34 @@ if(${GTEST_FOUND})
...
 
+    facelift_add_test(UnitTestsObserver
+        SOURCES FaceliftObserverTest.cpp
+        LINK_LIBRARIES
+            ${FACELIFT_GTEST_LIBRARIES}
+            FaceliftIPCCommonLib
+    )
 else()
-    message(WARNING "Required package google test not found!")
+    message(ERROR "Required package google test not found!")
```
bitmouse: WARNING!

tests/unittest/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+** CONDITION FOR PROCESSING: for the inspected object, the interface signal has changed, while the ready property is in the "ready" state. In this case, the functors in observers will be executed once.
+** If the interface signal for the inspected object has changed, but the ready property is in the "not ready" state. In this case, the functors in observers will not executed.
+** REQUIREMENT: One change in the interface signal and "ready" state corresponds to a single execution of the functor. No signal can be missed.
+** Main steps:
+** - Creation of main objects: object for inspection "Proxy" (this is one of more possible types), object for observation "IsReadyObserver", objects for executing "Observers".
+** - Installing objects/function to be executed in observers depending on the need: "callOnReady" or "callOnceReady".
+** - Setting up to "IsReadyObserver" to watch out for signal in "Proxy".
+** - Catching signal and execution of functors.
+** - Completion of work.
+**********************************************************************/
```
bitmouse: 1. Why you add something like this here? This is test, not a documentation. I have not checked the content, because it should not be here.
2. This formatting is like from 90'. Please remove unnecessary stars.

tests/unittest/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    ~IPCProxyTest() {}
+};
+
+TEST_F(IPCProxyTest, MinimalCoreTest)
+{
+    /**********************************************************************
+    ** INPUT: property of Proxy ready() in state disabled, trigger changeReady() is toggles 3 times.
+    ** OUTPUT: No observer should be executed.
+    **********************************************************************/
```
bitmouse: remove or make it readable

tests/unittest/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,336 @@
...
+    ASSERT_EQ(spyReadyObserver.count(), 0);
+
+    // Check values after signals call
+    ASSERT_EQ(c1.getValue(), 0); // for StandartObserver ("callOnReady")
+    ASSERT_EQ(c2.getValue(), 0); // for SingleTimeObserver ("callOnceReady")
+
+    /**********************************************************************
+    ** INPUT: property of Proxy ready() in state enabled, trigger changeReady() is toggles 3 times.
+    ** OUTPUT: All observer should be executed.
+    **********************************************************************/
```
bitmouse: remove

bitmouse (changes requested):
Please fix tests. I have not idea what they test, because you created two test cases within the test fixture, but it looks like there should be at least 10 of them. Please split them into separate TEST_F. 

I don't understand the point of IObserver and its virtual method. Could you explain it please?

tests/unittest/FaceliftObserverTest.cpp
zusage: Done

tests/unittest/FaceliftObserverTest.cpp
zusage: Done

tests/unittest/FaceliftObserverTest.cpp
zusage: Just for convenience. I did remove it.
Done

tests/unittest/CMakeLists.txt
zusage: Done

tests/unittest/CMakeLists.txt
zusage: Done

tests/unittest/FaceliftObserverTest.cpp
```diff
@@ -0,0 +1,273 @@
...
+    ASSERT_EQ(c1.getValue(), 0); // for StandartObserver ("callOnReady")
+    ASSERT_EQ(c2.getValue(), 0); // for SingleTimeObserver ("callOnceReady")
+}
+TEST_F(IPCProxyTest, proxyReadyIsTrue)
+{
+    /**
+     * @brief checking if proxy is ready and signal exists, all observer should be executed.
+     * @param set to Proxy in state "true"
+     * @param set 2 observers
+     */
```
bitmouse: this is not a place for doxygen documentation!!!!

zusage: Done

PR #317, opened 2020-10-22
Add QMetaEnum usage for autogenerated code

kunichik:


src/model/FaceliftEnum.cpp
```diff
@@ -32,10 +32,15 @@
 #include <QDebug>
 
 namespace facelift {
+    
+namespace Enum {
```
bitmouse, line 36: lowercase namespace please. This will allow us to distinguish it from classes

src/model/FaceliftEnum.h
```diff
@@ -31,11 +31,41 @@
...
+#include <type_traits>
+#include <memory>
+#include <QString>
+#include <QMetaEnum>
+#include <QByteArray>
 
 namespace facelift {
 
-void onAssignFromStringError(const QString &s);
+namespace Enum {
```
bitmouse, line 42: lowercase

src/model/FaceliftEnum.h
```diff
@@ -31,11 +31,41 @@
...
+    return QMetaEnum::fromType<T>().valueToKey(static_cast<int>(value));
+}
+
+// Returns the enumaration value of the given enumeration key, or nullptr if key is not defined.
+// TODO change std::unique_ptr to std::optional when it will be possible
+template<typename T>
+std::unique_ptr<T> fromString(const QString &string)
+{
+    QByteArray byteArray = string.toLocal8Bit();
+    bool ok {};
```
bitmouse, line 60: bool ok = false;
It is more explicit and if someone is just looking at the code, he will notice it faster than default bool value

src/model/FaceliftEnum.h
```diff
@@ -31,11 +31,41 @@
...
+    int value = QMetaEnum::fromType<T>().keyToValue(byteArray.data(), &ok);
+    
+    T result = static_cast<T>(value);
+
+    return ok ? std::make_unique<T>(result) : nullptr;
 }
+
+
+} // end namespace Enum
+
```
bitmouse, line 70: no need for empty line between namespace closing

bitmouse (changes requested):

src/model/FaceliftEnum.cpp
kunichik: lowercase name "enum" is reserved keyword.
Other names would be worst...

src/model/FaceliftEnum.h
kunichik: done

src/model/FaceliftEnum.h
kunichik: done, the same in cpp-file

src/model/FaceliftEnum.h
kunichik: let it be Enum as exception

src/model/FaceliftEnum.cpp
kunichik: let it be Enum as exception

codegen/facelift/templates/Enum.template.h
```diff
@@ -74,37 +74,20 @@ template<> inline const std::initializer_list<{{enum.fullyQualifiedCppType}}>& v
...
-    if (s == "{{member}}")
-        v = {{enum.fullyQualifiedCppType}}::{{member}};
-    else
-    {% endfor %}
-        ::facelift::onAssignFromStringError(s);
+    auto valuePointer = facelift::Enum::fromString<{{enum.fullyQualifiedCppType}}>(string);
+    if(!valuePointer) {
+        facelift::Enum::fromStringError(string);
+    }
+    value = *valuePointer;
```
bitmouse, line 86: UB?

kunichik: if there is no requested value in an enum, fromString returns nullptr inside the unique_ptr.
And is valuePointer holds nullptr, fromStringError raises qFatal.

Maybe it makes sense to move qFatal here ?

bitmouse: What was the action in this/similar case previously?

kunichik: Ok, and if someone will drop qFatal in the future we don't get into UB
will add else statement.

bitmouse (dismissed):

src/model/FaceliftEnum.cpp
```diff
@@ -32,10 +32,14 @@
...
```
jacky309, line 36: In facelift, namespaces are used to group entities which belong together into a layer. A upper-level layer (such as "facelift") should not have any dependency to a lower layer (such as "facelift::ipc"). Your change is breaking that rule.
Also, if you introduce a "Enum" namespace (whose name is also not right), you should also introduce a "Struct" namespace, but I do not see any benefit in doing so anyway.

src/model/FaceliftEnum.cpp
```diff
@@ -32,10 +32,14 @@
...
 
-void onAssignFromStringError(const QString &s)
+void fromStringError(const QString &string)
```
jacky309, line 38: bad function name. Either the function name indicates what the function does (with a verb), or it indicates when it is called (onXXXXX).

src/model/StringConversionHandler.h
```diff
@@ -65,77 +64,77 @@ struct StringConversionHandler<QVariant>
...
-            str << StringConversionHandler<ElementType>::toString(element);
+            str << StringConversionHandler<T>::toString(element);
             str << ", ";
         }
         str << "]";
         return s;
     }
 };
 
-template<typename ElementType>
```
jacky309, line 116: Why are you changing those names ??

src/model/FaceliftEnum.h
```diff
@@ -31,11 +31,40 @@
...
+template<class T, std::enable_if_t<QtPrivate::IsQEnumHelper<T>::Value, T>* = nullptr>
+QString toString(T value)
+{
+    return QMetaEnum::fromType<T>().valueToKey(static_cast<int>(value));
+}
+
+// Returns the enumaration value of the given enumeration key, or nullptr if key is not defined.
+// TODO change std::unique_ptr to std::optional when it will be possible
+template<typename T>
+std::unique_ptr<T> fromString(const QString &string)
```
jacky309, line 57: You are causing some data to be allocated on the heap, which is definitely not needed to convert a string to an enum.

jacky309 (changes requested):

src/model/FaceliftEnum.cpp
bitmouse: Sorry, but coding standards/guidelines in facelift, also community page, etc. are EMPTY. This means that you cannot expect any coding standards from anyone. 

Also separate namespace for utility makes sense and is a common practice in the software world, especially when you do not want to create a class/struct with all static methods.

src/model/StringConversionHandler.h
bitmouse: I don't know the reason, why changed names, but I like the new version more. It follows general practices in software for templates, so it should be easier to read for experienced user.

src/model/FaceliftEnum.cpp
kunichik: >In facelift, namespaces are used to group entities which belong together into a layer. A upper-level layer (such as "facelift") should not have any dependency to a lower layer (such as "facelift::ipc"). Your change is breaking that rule.

in this case namespace is used for its straight purpose - name scope.
Any other options like different name, class or structure are worst alternatives


>Also, if you introduce a "Enum" namespace (whose name is also not right), 

please suggest a better name

>you should also introduce a "Struct" namespace, but I do not see any benefit in doing so anyway.

it's not needed at the moment

src/model/StringConversionHandler.h
kunichik: yes it's not related to the current changes, but
 typename ElementType, typename Type, typename TypeName are bad in most cases and only T enough here and I just change this small change here.
We need a separate task for replacing  non common template parameters names and
hold something like ElementType only for rare cases when it's needed for better readability.

src/model/FaceliftEnum.h
kunichik: I considered that, only std::optional should be here.
an other options like struct, out parameter are worse and 
std::unique_ptr will be replaced with std::optional in the future

src/model/FaceliftEnum.cpp
kunichik: agree!
how about raiseFatalError ?

src/model/FaceliftEnum.cpp
kunichik: actually StringConversionHandler is not needed at all
and in the future it should be replaced with toString<class T> template function
with different specifications and SFINAE principle.
Also @bitmouse suggested the idea to move the enums stuff into a separate folder.

PR #324, opened 2020-11-02
Fix heap-use-after-free for IPCProxyBinderBase

zusage:


src/ipc/ipc-common/IPCProxyBinderBase.cpp
```diff
@@ -47,7 +47,7 @@ void IPCProxyBinderBase::connectToServer()
 
 bool IPCProxyBinderBase::isReadyToConnect() const
 {
-    return m_componentCompleted && enabled() && !objectPath().isEmpty();
+      return m_componentCompleted && enabled() && (!objectPath().isNull())?(!objectPath().isEmpty()):(false);
```
bitmouse: this is hard to read. Please split this into pieces, e.g. extract object path condition. Also please remember that C++ works from left to right, so you do not have to add there explict false, something like this is enough: ... && !objectPath().isNull() && !objectPath.isEmpty()...

bitmouse (changes requested):

src/ipc/ipc-common/IPCProxyBinderBase.cpp
zusage: Done.
I agree. Your variant is more elegant and readable.
Little improvement of your sentence "C++ works from left to right..."
Order of evaluation of any part of any expression, including order of evaluation of function arguments is unspecified, with some exceptions for the next: 
1) logical operators (&& and ||)
2) logical operator (?:)
3) comma operator ,

